PR #144, opened 2016-09-01
Rename con_tags to prim_tags

mn200:
Also make prim_tags obey more standard conventions


mn200:
Actually, there's a related change I want to make first.

PR #145, opened 2016-09-02
Rename con_tags to prim_tags

mn200:
Also make prim_tags obey more standard conventions

Sorry to be back-and-forthing with this; the subsequent change I want to make does depend on this, but it should be considered an independent "project".


xrchz:
You can add more stuff to a pull request without opening a new one, I think.

Issue #138, opened 2016-08-16
Optimise GetGlobal for Number constants in clos_known


myreen:
The `clos_known` optimisation tracks where values flow and can find that a `GetGlobal` will always return a specific number. In these cases it would make sense for `clos_known` to turn the `GetGlobal` into a `Const ...`. Such a change would allow BVL's constant folding routine, `bvl_const`, to make the code better.

mn200:
Cool idea.  I'll definitely take a look.

mn200:
Hmm. This is not quite so simple.  Calls to `SetGlobal` pass closures, not numbers, and similarly, the expected returns from `GetGlobal` calls are closure values. The `known` pass would have to record the actual `Fn` expression in its approximation map, and then substitute that expression for the `GetGlobal`. (And it would have to make sure not to recurse into that expression :-) Would that be worth doing?

myreen:
I'm a bit confused now. It's true that known is mostly interested in where closure values flow. However, it also tracks where integers flow.

How does the following example assign closures to globals?

    val c = 4;
    val z = c + 5;

I would have thought that the example above would become:

    SetGlobal 0 4 ;
    SetGlobal 1 (get global 0 + 5)

And would have thought that known knows that `GetGlobal 0` always returns `4`. Thus known could optimise the second line above to:

    SetGlobal (4 + 5)

Later in BVL, `4 + 5` gets collapsed to `9`.

In terms of code change to the compiler, this would be a very small change to the op case only: if the op is a `GetGlobal` and the returned approximation is some `Int`, then replace the op with `Op (Const ...) []`.

This is not exactly a major optimisation, but I hope it would also be easy to add.

mn200:
Ah!  I hadn't thought about just the Int values.

myreen:
By the way, as an extension: would it be difficult to make the approximations also track tags of conses/blocks, i.e. `Block tag ...`? A similar `GetGlobal` optimisation could be done for empty conses (such as `true`, `false` and `[]`). I would imagine that it would be helpful in removing dead code in case someone writes:

    val debug = false
    ...
    fun complicated_function ... = ... (if debug then print-lots-of-stuff else ()) ...

Currently the approximation tracks blocks/conses as `Tuple`. Maybe it would just be a matter of adding the `tag` number to the `Tuple` approximation. The optimisation could always replace a `GetGlobal` that loads a `Tuple tag []` with `Op (Cons tag) []`, for any value of `tag`.

PR #147, opened 2016-09-02
Simple data flow optimisations in clos_known

mn200:
The optimisations are as described in [the github issue](http://github.com/CakeML/cakeml/issues/138), and involve noticing helpfully precise approximations for the values returned by the `Global` op.

### DIFF
--- a/compiler/backend/clos_knownScript.sml
+++ b/compiler/backend/clos_knownScript.sml
@@ -111,14 +111,18 @@ val clos_gen_def = Define`
   (clos_gen n i [] = []) ∧
   (clos_gen n i ((x,_)::xs) = Clos (n+2*i) x::clos_gen n (i+1) xs)`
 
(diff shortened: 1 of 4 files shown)


xrchz:
I'm going to try using this as an example of regression testing before merging. Potential issues: that means waiting a long while (for the regression test to progress in the queue and then complete), and the results (on Bamboo) are not at all integrated with GitHub, so hard to see what's going on.

mn200:
Sure.  This one has the advantage that 

1. I don't suppose anything will want to build on it, in the short term at least; and
2. (equally), it shouldn't interfere with anything anybody else wants to do

xrchz:
Waiting for the Bamboo queue is too long, so I did a manual build of this merge up to `compilerProofTheory`; I consider that sufficient. For future, maybe when we get more machines (than just lemma) for building regressions using the automated system will be workable...

PR #153, opened 2016-09-20
Refactor/improve full normalise

Armael:
This redefines [full_normalise] in the style of CFML's. In particular, it takes care of pulling the introduced let-bindings, and not introducing too many, in order to avoid introducing nested let-bindings, which are annoying to deal with.

The termination proof is cheated for now. 
I'm not sure how morally acceptable it is, but I think using [full_normalise] to certify case studies is for now more important than proving termination.

To prove termination, I think one would have to merge the auxiliary definitions (`protect*`) into `norm`, not inlining them but having an integer argument to dispatch, then having as termination relation a lexicographic order of the expression size and the dispatching index...
(by dispatching I mean having 
```
norm mode other_args =
  if mode = 0 then (
    current code of norm
  ) else if mode = 1 then (
    current code of protect
  ) else if mode = 2 then (
    current code of protect_row
  ) ...
```
)

### DIFF
--- a/characteristic/cfComputeLib.sml
+++ b/characteristic/cfComputeLib.sml
@@ -23,17 +23,11 @@ val add_cf_aux_compset = computeLib.extend_compset
      cfTheory.pat_typechecks_def,
      cfTheory.pat_without_Pref_def,
      cfTheory.validate_pat_def,
-     cfTheory.build_cases_def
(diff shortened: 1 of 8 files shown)


Armael:
Woops indeed. Should be fixed now.

PR #154, opened 2016-09-24
Exceptions support in CF

Armael:
Not finished yet.

Progress:

- [x] Support `raise` and `handle` in the `cf` function
- [x] Update the soundness proof
- [x] Update the existing tactics and specifications of the examples so they still work
- [x] Implement tactics and add examples for `raise` and `handle`
- [x] Be able to use translator representation predicates for exceptions
- Optionally generalize existing cfs (eg for division or array access), to specify the case where an exception is raised (at the moment, the user has to prove that this case cannot happen) **[for later]**

Some questions:
- Does something need to be added to `basisProgramScript` about exceptions?
- please bikeshed about the auxiliary predicates used for post-conditions. At the moment, one writes `POSTv v. Qv v` for writing a "pure" postcondition (meaning that the exception case cannot happen), `POSTe v. Qe v` for a postcondition for the exception case only (the value case cannot happen), and `POST Qv Qe` for specifying one post-condition for each case
- does something else need to be done?

### DIFF
--- a/basis/basisProgScript.sml
+++ b/basis/basisProgScript.sml
@@ -104,18 +104,18 @@ fun prove_ref_spec op_name =
 
 val ref_spec = store_thm ("ref_spec",
   ``!xv. app (p:'ffi ffi_proj) ^(fetch_v "op ref" (basis_st())) [xv]
-          emp (\rv. rv ~~> xv)``,
+          emp (POSTv rv. rv ~~> xv)``,
   prove_ref_spec "op ref");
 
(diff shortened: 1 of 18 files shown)


xrchz:
Looking good. I think it's fine to just finish `raise` and `handle` tactics for this, and anything more can be done later.

Armael:
Everything should be okay now.
I'll rebase on top of master later tonight.

Armael:
Rebase done (it includes new changes, related to normalisation).

PR #168, opened 2016-11-10
Johannes

IlmariReissumies:
Miscellaneous cleanup work.

For example:
- Goals are stated as quotations rather than terms, as per the convention
- Semantics of all languages updated to use fix_clock rather than check_clock style
- replace_term -- a curious substitution function that does not respect alpha-equivalence and may cause variable capture -- removed


xrchz:
A regression test for this is currently running.

xrchz:
It mostly succeeded - there were some `rator_x_assum`s remaining in the bootstrap that made it fail there. But I'll fix those up on master after merging.

PR #169, opened 2016-11-12
Make memory loads a little bit more efficient 

myreen:
The default is to do two shifts to extract the address from pointer
with information in the lower bits. This commit makes the compiler
produce code with only one shift if there are enough pad bits. There
needs to be two pad bits for 64-bit archs and one pad bit for 32-bit
archs (because the actual number of pad bits is 1+pad_bits).


tanyongkiam:
Should the defaults in configTheory be updated to what you mention here? (I'll make the change once this is in master)

myreen:
@tanyongkiam yes, that sounds like a good plan.

PR #185, opened 2016-11-25
Adjust gen_readme to warn, not die, on "bad" dirs

mn200:
Also remove generated README.md files from repo.


xrchz:
Just checking a few things here:
- Are the `README.md` files left checked in and not cleaned? They should be (so they appear on GitHub).
- Does the tool complain about undocumented directories? I think we want this behaviour to encourage a high standard of documentation.

mn200:
The `Holmakefile` for the root directory doesn't clean the `README.md` file; the one in `developers` does. This is clearly not right. My feeling is that it's reasonable to have them all cleaned as they are trivial to regenerate (contrast the situation with `lem`-generated script files). They are certainly now both back in the repo (in afa7436).

The tool does indeed complain about directories with no documentation.  It doesn't fail though, and you won't see the warning message if you're running parallel Holmake because it just goes into the `.hollogs` file.

xrchz:
I think it should fail for undocumented directories, because it's easy to add a `readmePrefix` file to stop the failure (even if it just says "this directory is undocumented").

The problem with cleaning things in the repository is that `Holmake clean` can change your `git status` to "modified" - that is likely to be annoying/surprising.

mn200:
OK.

PR #192, opened 2016-12-05
Make progress on data_to_word, new instructions in asm, Equal etc.

myreen:


xrchz:
The build is currently failing... on the builds list this branch is the one where the path include `CML-RT15`. (Making this more obvious is #194)

tanyongkiam:
Noting before I forget: the unverified allocator must be updated for forced edges on MIPS and RISCV.

PR #205, opened 2016-12-13
Basis

ConnorCashman:


xrchz:
This currently has a dependency circularity: `translator` depends on `characteristic` and vice versa. `translator` should not depend on `characteristic`. Fix that up and then we'll see if the regression works...

Issue #232, opened 2017-03-06
Stop using match_goal in CF tactics


Armael:

Armael:
Closed by #233 .

PR #233, opened 2017-03-06
Stop using [match_goal] in the implementation of CF tactics

Armael:
Use helpers from cfSyntax and cfAppSyntax instead.
This should make the tactics a little more robust wrt parsing-related changes as they don't use quotations anymore.


Armael:
This fixes #232 .

Issue #227, opened 2017-02-22
Avoid general-purpose comparisons inside compiler


myreen:
Some performance was lost when bignums were introduced. The code for addition and subtraction is likely to be of similar performance as before; multiplication and division were not available before so they can't be the reason. This leaves the comparison operations as the main suspect.

If a user writes `n < k` in their programs, then I think we can't do much to improve the quality of the code without some potentially heavy static analysis (and luck).

However, there are places where the compiler will generate `<` internally. I can think of two places:
 1. bounds checking code for array and vector accesses
 2. the jump tables in the `clos_to_bvl` stubs

Both of these could be compiled to better code in `data_to_word` if there were special ops in ClosLang that support these as primitives. I'm thinking that we should have:
 1. `BoundsCheck` and `BoundsCheckByte` primitives introduced in `pat_to_clos`
 2. `SmallLessConst` introduced in `bvl_jump`

xrchz:
Isn't this the same as #79?

myreen:
You are right. I'll close this one and refer to #79 in commits related to floating-point.

PR #237, opened 2017-03-09
Improve compiler generated bounds checks (#227)

myreen:
A regression test for this pull request is running on chit.


myreen:
The regression test has now passed all the crucial points.

PR #263, opened 2017-04-23
Cleanup

xrchz:
Some minor improvements. I hope to do more later.


mn200:
I thought the parsing tests were already working.  Certainly they were when I adjusted the PEG soundness and completeness proofs.

xrchz:
They would have been broken by 7fbd1d01861385d3298aa5e6ad147a6de2b69f48

mn200:
Ah, right.  I forgot that declarations picked up locations too.

PR #270, opened 2017-04-26
Add generational GC

myreen:


xrchz:
Running a regression test on this

xrchz:
Compiler evaluation (starting with `compiler/eval/benchmarks`) seems to go into an infinite loop.

PR #275, opened 2017-05-14
Fix tail calls for lets in BVI

xrchz:
@tanyongkiam is this ready to go? backend proofs built ok.


tanyongkiam:
I don't think it'll break anything else, but it might be nice to get a full
build done and compare the performance of this with a recent bootstrap
before merging.

Issue #276, opened 2017-05-16
Make translator directly support w2w


myreen:
This was requested by @tanyongkiam.

xrchz:
Closed in #277

PR #277, opened 2017-05-16
Add direct translator support for w2w

myreen:
For #276


xrchz:
regression test is running for this. (let me know if it's already tested.)

PR #279, opened 2017-05-22
Implement the cf normalise pass as a ML function

Armael:
To do:
- [x] - measure if there's an actual speed-up compared to evaluation in the logic
- [x] - do some clean-up (disconnect the old normalise in-logic definition)
- [x] - write some comments


Armael:
I hadn't realized how slow the HOL normaliser was on big examples. Running the HOL normaliser on `partition` from `characteristic/examples/quicksortProgScript.sml` takes 2 minutes 40 on my computer, while the ML normaliser terminates almost instantly.

Armael:
I think this is good to go!

PR #287, opened 2017-06-12
Haskell constructors

mn200:
I think this is ready for merge.  Soundness and completeness proofs have been fixed.


xrchz:
I'm running a regression on it.

mn200:
The `basis` directory builds successfully on my machine as of 1245d9a

xrchz:
Is this replaced by the `haskell-datatype-decl` branch? I think this branch shouldn't be merged because it breaks `characteristic/examples` (due to `ref`) but maybe that's being fixed on the other branch.

mn200:
Yes.  I'll fast forward this branch to be the same as the other to make things easier though.

mn200:
Latest Bamboo regression looks to be timing out in `characteristic/examples/compilation/proofs` in the same way as happened on master.

PR #294, opened 2017-06-27
Introducing tail recursion in BVI

oskarabrahamsson:
The proofs are cheat-free and I think that this is ready for regression testing.


xrchz:
Regression test is in the queue, hopefully should be finished in a couple of days.

PR #295, opened 2017-06-29
Improve init of the TriggerGC pointer

myreen:


xrchz:
Regression test in queue.

PR #296, opened 2017-07-03
Make bignum machinery allow non-tail-recursive functions

myreen:
This is needed for a MSc project, which aims to implement 
Karatsuba multiplication for CakeML's bignum library.


xrchz:
regression test is queued on bamboo

PR #299, opened 2017-07-08
Swap tag values for true and false

### LINKED ISSUE #298: Swap tag values for true and false
Currently, `true` is represented as `Block 0 []` and `false` as `Block 1 []`. These ought to be the other way around as the current representations cause sub-optimal code in `data_to_word`.

xrchz:
Closes #298


myreen:
I think the `issue298` branch is ready for a full regression test and then merge. The compiler proofs go through.

PR #305, opened 2017-08-25
Add ocaml floating-point benchmarks

malyzajko:


myreen:
Thanks @malyzajko. These small examples will work nicely as target examples for CakeML's float support. I think the hardest missing part is parsing of float literals (into word64s).

PR #310, opened 2017-09-04
bvi_tailrec contexts

oskarabrahamsson:
Extends the analysis pass in the bvi_tailrec phase with some conservative type analysis that enables the compiler to figure out the types of variables in certain instances.

For instance, the following rewrite is now possible (since `n` can only be lifted into the accumulator if we know beforehand that it evaluates to an integer):
```
fun fac n = if n <= 1 then 1 else fac (n-1) * n
===>
fun fac' n acc = if n <= 1 then 1*acc else fac' (n-1) (n*acc)
```


xrchz:
As you may have seen, this appears to break the bootstrap translation in `to_dataProg`

oskarabrahamsson:
I eliminated some sources of `ARB` and sorted out the translation of `bvi_tailrecTheory`-things in `to_dataProg`, and managed to get the translation all the way through the BVI phase (past my additions).

PR #313, opened 2017-09-11
2-argument FFI calls

IlmariReissumies:
Make FFI calls take two arguments -- one immutable string and one mutable bytearray --- rather than one as previously.

@tanyongkiam I tagged you as suggested reviewer because I've changed the register mappings for most architectures (to make them agree with C calling conventions in the first 4 args), but I have only tried out the generated binaries for x86. It would be reassuring if you could check that I haven't done anything obviously insane, and perhaps even check that the generated binaries for other architectures do not segfault :)


xrchz:
The regression test seems still to be stuck around `mips_configProof` - let me know if you think it actually gets past this.

PR #315, opened 2017-09-18
Fix contexts in bvi_tailrec

oskarabrahamsson:
This fixes an issue with the analysis pass in `bvi_tailrec` losing track of variable types when traversing `Let`-expressions, causing the optimization to not be applied in all cases that it should. 

In addition, two theorems from `bvi_tailrecProof` that I originally copied from `bvlProps` are moved to `bviProps`.


xrchz:
regression test is running. I wonder how badly this will conflict with `install-and-run`...

oskarabrahamsson:
I was hoping that it would not interfere too much; very little happened in the `evaluate_rewrite_tail` theorem at least.

xrchz:
bamboo seems to be stuck at `compiler/backend/arm6/proofs` for some reason. could this be a real problem, or should I just reset it?

oskarabrahamsson:
That sounds dubious, I think you should try to reset it.

PR #318, opened 2017-09-20
Fuse remove annotate

xrchz:


xrchz:
I got the bootstrap translation through to `explorerProg` on my laptop, as well as the benchmarks. (As well as the compiler proofs of course.) So I think any remaining problems can be resolved on master.

PR #329, opened 2017-10-05
Fileioffi

hferee:


xrchz:
I think the `ltemporal` and `optionMonad` theories should simply be added to HOL, probably merged into `llist` and `option`. You could put them into `misc` for the time being.

hferee:
Do you mean the misc Script or directory?

xrchz:
I mean `miscTheory`.

xrchz:
This is failing the regression test rather early (in `basis`)

hferee:
I'm running the regression test on a fresh install and it went through basis with no issue (although it might fail later).

hferee:
However, it fails in `compiler/backend/gc` because of an overloaded notation. I'll fix it and re-run the regression test before pushing.

hferee:
This should be fine now.

xrchz:
I expect it will still fail on basis: what needs to be tested is not this branch as it is, but its merger with master (since that's what the new master will become).

PR #348, opened 2017-10-22
Monad trans

Kachoc:
Improve the precondition generation for the monadic translator by lifting the monadic state as a parameter of the EvalM predicate. Improve the m_translate_run mechanism to output shorter theorems to make debugging easier (application of simplifications, introduction of abbreviations).


xrchz:
(this is in the regression test queue)

Issue #326, opened 2017-09-28
Make the bootstrapped compiler a cross-compiler


xrchz:
This issue is to add compiler command-line flags to the bootstrapped compiler which let the user choose the compilation target. Most of the bootstrap translation is shared: only the encoders differ, at least for the 64-bit targets. For this issue, 32-bit ARM can be left out. The compiler that takes the target flags should then be bootstrapped for each target.

tanyongkiam:
This is closed by #352 (for the 64-bit targets):
./cake --targetx64 < foo.sml > foo_x64.S
./cake --targetarm8 < foo.sml > foo_arm8.S
./cake --targetmips < foo.sml > foo_mips.S
./cake --targetriscv < foo.sml > foo_riscv.S

PR #352, opened 2017-10-25
Arm6 (and 8) export

tanyongkiam:
This updates the arm6/8 exporters and bootstraps a cross compiler from x64 to both targets.


xrchz:
Can you update `build-sequence`?

xrchz:
Also the naming conventions are inconsistent here. Observe that `compiler/bootstrap/compilation/riscv/riscvBootstrapTheory` produces a `riscv`-outputting compiler for `riscv`. But you've put an `arm6`-outputting compiler for `x64` in the analogous theory location for `arm6`.

xrchz:
we figured some of this out on chat. pasting here:

`bootstrap/translation` culminates in two translations `compiler64Prog` and `compiler32Prog` - the first ends with a CF-verified `main` program for the CakeML compiler for all 64-bit out-targets, and the second is the same for arm6 as the out-target. N.B. the current `compiler64Prog` theory does not do this, but I suspect it should be extended in order to do so. Then delete the subdirectories of `translation`. Optionally replace the `32` and `64` suffixes on some of the `Prog` theories with subdirectories by those names instead.


for `bootstrap/compilation` the subdirectories look like `run-target/{32,64}/proofs`; within the 32/64 directory is a `runtargetBootstrap` theory that produces the compilation theorem (and outputs the `.S`)

xrchz:
I won't run the regression while the requested updates above are still in progress.

xrchz:
Regressions are now running automatically since this pull request is open. Is it actually wanting to be merged though?

tanyongkiam:
I'll close the PR

tanyongkiam:
I'm re-opening this because I probably can't run the bootstrap to completion on my machine (runs of of stack).

This implements the 64-bit version of #326 for x64. The riscv/arm8/mips versions can be similarly added if desired. It is unclear if the mips export script is actually correct.

Note that this includes PR #377 because I did not want to clash with that branch after it gets merged.

xrchz:
What happens to the artefacts (tarballs containing bootstrapped compiler) after these changes? Do they have a new location?

tanyongkiam:
I didn't test that, but I assumed it will be generated as usual by the Holmakefile here in this directory as cake-x64:
https://github.com/CakeML/cakeml/tree/arm6_export/compiler/bootstrap/compilation/x64/64

xrchz:
Hmm, the symlinks there look incorrect - they probably need an extra `..` since they are one directory deeper now.

xrchz:
Do you have `bootstrap/translation` in the build sequence twice?

PR #364, opened 2017-11-03
Regression test tweaks

xrchz:


xrchz:
The regression test for this died when building HOL due to HOL-Theorem-Prover/HOL#461

xrchz:
I'm abandoning these changes, since alternative regression infrastructure seems to work better.

PR #366, opened 2017-11-07
Compilation of 32-bit versions of WordFromInt and WordToInt

myreen:
These used to be implemented as `GiveUp` for 32-bit architectures. Now they have real implementations.


xrchz:
I'm still working on sending testing status updates (both to GitHub and via email). But there is a failed regression test for this with output here: https://cakeml.org/regression.cgi/job/6
(If new commits are pushed to this pull request, it will be tested with a new job number.)

myreen:
Thanks! I'll fix this.

Issue #373, opened 2017-11-13
Include commit hash in bootstrapped compiler


myreen:
It would be good if the bootstrapped compiler had a command-line option (e.g. `--version`) which prints what git commit this bootstrapped compiler was built from.

This could be implemented by including an ML-generated term in some constant definition `current_version`, which gets translated like any other constant. The ML code would ask the shell what git commit the checked out code is at. 

What would the right git command be?

Should this fetching of the current git hash be functionality provided by HOL4? I can imagine that it might be useful in projects other than CakeML. If the HOL4 library wants to be fancy, then it could print what PolyML is currently used, what HOL commit is current, and the commit of the current directory (somewhere in CakeML for us).

xrchz:
A git command to get the current head commit is `git rev-parse HEAD`.

oskarabrahamsson:
This functionality got merged into master with #352.

PR #377, opened 2017-11-14
Include commit hash in bootstrapped compiler

oskarabrahamsson:
This pull request addresses #373. 

A `--version` flag is added to the bootstrapped compiler binary, which shows the following (when available) before terminating.

* Build date
* The CakeML commit hash
* The HOL4 commit hash
* PolyML version


oskarabrahamsson:
I added some functionality to preamble:

```sml
(* tm_from_proc cmd args *)
val tm_from_proc : string * string list -> term

(* tm_from_proc_from dir cmd args *)
val tm_from_proc_from : string * string * string list -> term

(* mlstring_from_proc dir cmd args *)
val mlstring_from_proc : string * string list -> term

(* mlstring_from_proc_from dir cmd args *)
val mlstring_from_proc_from : string * string * string list -> term
```

These all execute an external process and capture its standard output as string or mlstring terms. Optionally a working directory for the process can be specified with `xxx_from_proc_from`.

xrchz:
As you can see, the regression test failed.

xrchz:
Reopen when the failure has been addressed

oskarabrahamsson:
I ran the x64 bootstrap compilation myself this time. In addition to fixing the previous regression test failure, the `--version` flag is checked in the ML wrapper around the `compile_to_bytes` function (this is the only way to avoid reading input from `stdin`).

oskarabrahamsson:
Closing this as it is included in #352.

Issue #342, opened 2017-10-17
Nat.toString and Nat.fromString


xrchz:
- Define variants of the `toString` and `fromString` functions in `basis/pure/mlintTheory` that work on natural numbers rather than integers. (They should call the existing helper functions.)
- Prove them correct and add them to the basis library (either in `Int` or perhaps better in a new module `Nat`or `Num`). 
- Use these throughout the compiler and examples, in favour of any other number conversion functions.

agomezl:
I finish writing `basis/pure/mlnumTheory` and now I'm fiddling with `basis/NumProgTheory` but am not quite sure how to get it right.

I am currently "copying" what is in `IntProgTheory` but dont know what to do with:

```sml
(* What is this BTW? *)
val _ = ml_prog_update (add_dec ``Dtabbrev unknown_loc [] "num" (Tapp [] TC_num)`` I);

val _ = trans "+" `(+):num->num->num`
val _ = trans "-" `(-):num->num->num`
val _ = trans "*" `($*):num->num->num `

(* Should I try to these? *)
(* val _ = trans "div" `(/):num->num->num` *)
(* val _ = trans "mod" `(%):num->num->num` *)
(* val _ = trans "<" `(<):num->num->bool` *)
(* val _ = trans ">" `(>):num->num->bool` *)
(* val _ = trans "<=" `(<=):num->num->bool` *)
(* val _ = trans ">=" `(>=):num->num->bool` *)

```
is this necessary for Num? for the most part I understand what is happening but don't really know if this should be (or is already) done somewhere else.

Additionally,  when translating `toString` `fromString` and `fromString_unsafe` I get some inconsistent
naming from the translator.

```sml
(* Gives me tostring_v_thm *)
val result = translate (toString_def |> REWRITE_RULE [mlintTheory.maxSmall_DEC_def]);
(*
> Translating tostring
val result =
   |- (NUM --> STRING_TYPE) toString tostring_v:
   thm
*)

(* Gives me fromstring_unsafe_1_v_thm *)
val result = translate fromString_unsafe_def;
(*
> Translating fromstring_unsafe_1

WARNING: fromstring_unsafe_1 has a precondition.

val result =
    [.]
|- (Eq STRING_TYPE v1 --> NUM) fromString_unsafe fromstring_unsafe_1_v:
   thm
*)

(* Gives me fromstring_v_thm *)
val result = translate fromString_def;
(*
Translating fromstring

WARNING: fromstring has a precondition.

val result =
[...]

PR #379, opened 2017-11-16
Add Num.fromString and Num.toString

agomezl:
Resolves: #342


xrchz:
Reopen when updated. Also, I think the third bullet on #342 is not yet addressed.

agomezl:
@xrchz the dependency-order should be:
```
IntProgTheory -> NumProgTheory -> RatProgTheory
mlintTheory -> mlnumTheory
```
?

xrchz:
`mlint` and `mlnum` shouln't appear in the `dependency-order` file

PR #402, opened 2017-12-05
Add coalescing info to oracle allocator

tanyongkiam:
Adds a flag (default on) which makes the (unverified) allocator print out additional information about the code table entries.

For each code table entry, we get 1) the number of moves 2) the moves that don't clash and 3) the number of moves that got coalesced.


xrchz:
This seems to send the bootstrap translation into a loop

tanyongkiam:
It seems to have got past that point after @xrchz restarted the worker.
I don't see how my changes could have affected the translation so it might just be a one-off thing.

Issue #344, opened 2017-10-18
Automatically generate basis library contents listing


xrchz:
It is hard to know what's in the CakeML basis library without delving deep into the theories in `basis`; even if you know some functionality exists, its name and argument order might not be obvious. However, all the information is there and could be extracted into a presentable format.

This issue is to make producing a listing of basis library contents part of the normal build process (i.e., when `basisProgTheory` is built, it should also produce this listing). At minimum this listing should include the names and types of the library functions. It could also include further useful information such as corresponding HOL function (if any) and statement of CF specification (if any).

Some code that could be used as a starting point:
```
val _ = translation_extends"basisProg"
val () = Globals.max_print_depth := 20;

val get_dec_names_def = Define`
  get_dec_names (Dlet _ p _) = pat_bindings p [] ∧
  get_dec_names (Dletrec _ funs) = MAP FST funs ∧
  get_dec_names _ = []`;

val get_top_names_def = Define`
  get_top_names (Tdec d) = get_dec_names d ∧
  get_top_names (Tmod mn _ ds) = MAP ((++) (mn++".")) (FLAT (MAP get_dec_names ds))`;

val get_prog_names_def = Define`
  get_prog_names prog = FLAT (MAP get_top_names prog)`;

val prog_tm =
  get_ml_prog_state() |> remove_snocs |> clean_state |> get_thm
  |> concl |> strip_comb |> #2 |> el 3

val names =
  EVAL``get_prog_names ^prog_tm`` |> rconc
  |> listSyntax.dest_list |> #1
  |> map stringSyntax.fromHOLstring

app (fn x => (print x; print"\n")) names
```

#152 is vaguely related.

The reference manual might also want to include something like this.

myreen:
Each basis module should probably have a signature that hides all of the auxiliary functions, e.g. the accumulator versions of various functions.

I think `ml_progLib` should have a feature for attaching signatures to modules when closing them. This signature producing facility could force the user to provide the name, type and a textual documentation string for each externally visible binding. The documentation string would not appear in the `ml_progLib` generated AST. However, `ml_progLib` could, as a separate side-effect, generate documentation from the information it received when creating the signatures.

The reference manual should include an auto-generated section based on the signatures + documentation string of the basis modules.

lorenzleutgeb:
I think this issue was inappropriately closed automatically, looking at the commit message of 66a3531.

michaelsproul:
[...]

PR #404, opened 2017-12-09
Make basis know ^ (string concat.) in String and at top-level

mn200:
Show that one can reason with it in cf_examples, though xlet_auto
doesn't cope at the crucial moment for reasons that are unclear to me.


mn200:
`strcat` was already there so I didn't want to remove it.  But I agree that it seems superfluous.

xrchz:
I think we can do cleanup of unnecessary library contents after #344 is finished, so won't block this PR on that.

PR #406, opened 2017-12-11
Avoid parsing

xrchz:
Remove unnecessary calls to the parser in the monadic translator libraries.


xrchz:
I'm closing this while it's failing the regression test. Re-open when fixed.

PR #408, opened 2017-12-11
Issue 407

myreen:


xrchz:
I think x64BootstrapProof fails here, according to https://cakeml.org/regression.cgi/job/110.

xrchz:
Still failing. Reopen when fixed.

PR #409, opened 2017-12-14
Son's monadic translation of the inferencer

myreen:


xrchz:
The regression test looks like it may be in an infinite loop

xrchz:
Bootstrap translation is broken. Reopen when fixed. I might have a look.

xrchz:
This seems to work interactively but not non-interactively, so another case of load-order non-confluence.

xrchz:
Tracked this down to parsing in `cfLetAutoLib.sml` -- in particular, ``EQ`` gets parsed as a constant (from `ml_monad_translator`) or as a variable, depending on the load order.

xrchz:
Hopefully this will be fixed after #419 is merged

PR #410, opened 2017-12-19
MLton benchmarks

tanyongkiam:
This moves the MLton benchmarks into its own subdirectory, and adds both CakeML and SML versions.

In addition, it uses the bootstrapped multi-target compiler to build the benchmarks for a variety of purposes (GC testing, optimisation testing, target testing)


xrchz:
I had a quick look - this seems good. Does it actually work (i.e., you tried building them)? And, does `Holmake` in the CakeML root directory work properly (it was configured to run the benchmarks before)?

xrchz:
Closing pending reply from @tanyongkiam

PR #413, opened 2017-12-21
Mlrat as real

mn200:
Proved the homomorphism theorems that Magnus left for me.


xrchz:
Regression test failed - reopen when fixed

mn200:
Have pushed necessary prerequisite commit to HOL (HOL-Theorem-Prover/HOL@17dce4e3a006) - so reopening.

PR #417, opened 2017-12-30
Add support for compiling float operations to x86-64

myreen:


xrchz:
This breaks the bootstrap. Reopen when fixed.

tanyongkiam:
bootstrap/translation now builds

PR #420, opened 2017-12-31
Whole prog spec

xrchz:
Clean the way whole verified programs are created, making the `basis_ffiLib` automation less general.


xrchz:
Breaks regression. Reopen when fixed.

xrchz:
Bootstrap needs updating.

xrchz:
Bootstrap proofs need fixing

PR #422, opened 2018-01-05
Fixed isSubString so that it finds suffixes.

kifan-xiao:
For example, isSubString "a" "a" and isSubString "b" "ab" both returned
false previously.

### DIFF
--- a/basis/pure/mlstringScript.sml
+++ b/basis/pure/mlstringScript.sml
@@ -486,7 +486,7 @@ val isSubstring_aux_def = Define`
 val isSubstring_def = Define`
   isSubstring s1 s2 =
   if strlen s1 <= strlen s2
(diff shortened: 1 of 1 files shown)


xrchz:
Thanks! Would it be possible to also prove a theorem `isSubstring s1 s2 <=> IS_SUBLIST (explode s2) (explode s1)` to sanity-check this definition?

mn200:
Indeed, we hope to prove this result.

xrchz:
Regression test failed. Reopen when fixed (perhaps also after adding the proof).

xrchz:
Any update on this?

mn200:
It's done.   (See #455 and 82b7fd0926.)

xrchz:
Oh right - sorry! Thanks.

PR #426, opened 2018-01-09
 Support more leading chars in Int.fromString

### LINKED ISSUE #397: Support more leading chars in Int.fromString
Currently our `Int.fromString` only allows a leading "~" to indicate a negative integer. The [SML basis documentation](http://sml-family.org/Basis/integer.html#SIG:INTEGER.fromString:VAL) indicates that "-" and "+" should also be supported. This issue is to do so.

michaelsproul:
Closes #397

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,9 @@ timing.log
 # Emacs backup files
 *~
 
+# Vim swap files
+*.swp
+
 # Reference manual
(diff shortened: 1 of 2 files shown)


michaelsproul:
Yeah, I just have a habit of leaving vim open while I commit... dirty I know!

PR #434, opened 2018-01-17
Generate ML signatures for HOL functions in basis

michaelsproul:
This commit generates CakeML signatures for the HOL functions in basis, and stores them in the current translation by passing them into `close_module`.

The signature generation is turned on for each relevant module using ml_translateLib's `generate_sigs` flag.

The generated signatures are saved as definitions of the following form:

```
<cakeML name>_sig = Sval ...
```

For example, for a HOL function called `get_opt` that is translated to
`getOpt`, the theorem storing the signature looks like this:

```
getOpt_sig = Sval ... : thm
```

Some care is also taken to translate operator names into names appropriate for definitions, e.g. `div_sig` instead of `/_sig`.

Partially addresses #344 (still need a way to get signatures for pure CakeML code)

### DIFF
--- a/translator/ml_translatorLib.sml
+++ b/translator/ml_translatorLib.sml
@@ -2713,6 +2713,35 @@ fun dest_word_shift tm =
   if wordsSyntax.is_word_ror tm then Eval_word_ror else
     failwith("not a word shift")
 
+(* CakeML signature generation and manipulation *)
+val generate_sigs = ref false;
+
+fun sig_of_mlname name = definition (ml_progLib.pick_name name ^ "_sig") |> concl |> rhs;
+
+fun module_signatures names = listSyntax.mk_list(map sig_of_mlname names, spec_ty);
+
+fun sig_of_const cake_name tm =
+  mk_Sval (stringSyntax.fromMLstring (ml_progLib.pick_name cake_name), type2t (type_of tm));
+
+fun generate_sig_thms results = let
+  fun const_from_def th = th |> concl |> strip_conj |> hd |> strip_forall |> #2
+                             |> dest_eq |> #1 |> strip_comb |> #1;
+
+  fun mk_sig_thm sval = let
+    val cake_name = dest_Sval sval |> #1 |> fromHOLstring;
+    val sig_const_nm = cake_name ^ "_sig";
+    val sig_const_tm = mk_var(sig_const_nm, spec_ty);
+
+    val def = new_definition(sig_const_nm, mk_eq(sig_const_tm, sval));
+    in def
+  end
+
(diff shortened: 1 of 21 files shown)


michaelsproul:
I plan to take out the `FIXME`s before this PR is merged, I just wanted to seek your input on those points

translator/ml_translatorLib.sml
```diff
@@ -2713,6 +2713,35 @@ fun dest_word_shift tm =
   if wordsSyntax.is_word_ror tm then Eval_word_ror else
     failwith("not a word shift")
 
+(* CakeML signature generation and manipulation *)
+val generate_sigs = ref false;
+
+fun sig_of_mlname name = definition (ml_progLib.pick_name name ^ "_sig") |> concl |> rhs;
+
+fun module_signatures names = listSyntax.mk_list(map sig_of_mlname names, ``:spec``);
```
xrchz: Can you replace ``:spec`` with something that does not call the parser? Maybe that type is already in `astSyntax` or something?

translator/ml_translatorLib.sml
```diff
@@ -2713,6 +2713,35 @@ fun dest_word_shift tm =
...
+  mk_Sval (stringSyntax.fromMLstring (ml_progLib.pick_name cake_name), type2t (type_of tm));
+
+fun generate_sig_thms results = let
+  fun const_from_def th = th |> concl |> strip_conj |> hd |> strip_forall |> #2
+                             |> dest_eq |> #1 |> strip_comb |> #1;
+
+  fun mk_sig_thm sval = let
+    val cake_name = dest_Sval sval |> #1 |> fromHOLstring;
+    val sig_const_nm = cake_name ^ "_sig";
+    val sig_const_tm = mk_var(sig_const_nm, ``:spec``);
```
xrchz: same here

translator/ml_translatorLib.sml
michaelsproul: sure thing! I'll get onto that tomorrow :)

xrchz:
Fix the regression failure also please

mn200:
Note that it is possible to use the parser in library files safely, but that it is necessary to fix the grammars used.  See for example [the top of this library file](https://github.com/HOL-Theorem-Prover/HOL/blob/6c85649c3923843e53fb47a857db8a30b586873c/examples/l3-machine-code/m0/step/m0_stepLib.sml), where `structure Parse` is rebound.

michaelsproul:
Ok, I think I've addressed all the issues. I ppted for a `spec_ty` definition in `astSyntax`, and I _think_ I got all the import issues that were breaking the build \*fingers crossed\*

Issue #429, opened 2018-01-10
Use 8 byte file descriptors


myreen:
Currently, the way the basis talks to `basis_ffi.c` is to pass one byte as the file descriptor, because that is convenient with CakeML's FFI interface. However, the one-byte representation leads to strange limits on the number of possible files and also an ugly indirection in `basis_ffi.c`.

This issue can be closed when:
  - the CakeML basis uses 8 bytes to communicate file descriptors
  - the indirection is removed completely from `basis_ffi.c`

Ideally, the file descriptors would be:
  - represented as immutable strings (of length 8) in CakeML
  - passed as part of the constant argument in FFI calls

PR #443, opened 2018-01-24
Marshalling of integers in basis_ffi

hferee:
- file descriptors are encoded on 8 bytes (fixes issue #429 )
  - they are passed to the ffi via the new ffi string argument
  - as a consequence, the infds array in basis_ffi.c is removed
  - a constant (maxFD) is introduced to have an explicitly assumption on how many files can be simultaneously opened.
- read/write integer arguments/outputs are encoded on 2 bytes
  - the IO buffer size is now 2048 (+4 for arguments) instead of 256 (it can be increased up to 256^2 but I'm nut sure this is necessary/wanted)
  - after a quick benchmark, iocat's performance seem much closer to GNU cat now


xrchz:
currently this breaks the bootstrap translation - reopen when fixed

hferee:
These commits fixed this failure, but I'm still waiting for my local compilation to finish.

PR #451, opened 2018-02-04
Remove num_exp from WordLang

ncough:
WordLang's `num_exp` was used to represent calculations based
solely on constants, however, it was never used beyond wrapping
`num` values. This change removes `num_exp` and replaces its
uses with `num` directly.

### DIFF
--- a/compiler/backend/backendComputeLib.sml
+++ b/compiler/backend/backendComputeLib.sml
@@ -417,8 +417,7 @@ val add_backend_compset = computeLib.extend_compset
     ]
   ,computeLib.Tys
     [ (* wordLang *)
-     ``:'a wordLang$num_exp``
-    ,``:'a wordLang$exp``
+     ``:'a wordLang$exp``
     ,``:'a wordLang$prog``
       (* word_bignum *)
     ,``:word_bignum$address``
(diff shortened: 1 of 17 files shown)


xrchz:
The automatic regression test machinery doesn't handle pull requests - you may need @michaelsproul to add your repo to one of the worker machines then test this. (And/or add PR support to the regression code.)

michaelsproul:
I did some horrible things to the regression worker's git remotes on `cakeml1791`, and now it's running a build for this here: https://cakeml.org/regression.cgi/job/207

mn200:
Aren't all pull requests available at `origin/pr/ddd`?

michaelsproul:
@mn200 I don't think so, I don't see anything when I try to do `git checkout origin/pr/x`, and if you just have this repo cloned and try to check out Nick's commit, you get:

```
$ git checkout ca1328bad4fdadd20a276afc46ebf72df6256e88
fatal: reference is not a tree: ca1328bad4fdadd20a276afc46ebf72df6256e88
```

ncough:
According to this: https://help.github.com/articles/checking-out-pull-requests-locally/ its possible to fetch from `refs/pull/ddd/head`. I'll try to modify the regression code.

xrchz:
Awesome - regression testing PRs worked! And this PR failed the test, so reopen when fixed.

PR #452, opened 2018-02-05
Rename graphStateProg to floyd_warshallProg

myreen:


xrchz:
This doesn't make sense to test while `translator/monadic/examples` is not in the build sequence. Wasn't it supposed to come back?

PR #457, opened 2018-02-18
mlstring instead of string on failure for mlstring functions in preamble

tanyongkiam:
The stub that generates the version text in compilerScript may break otherwise.


oskarabrahamsson:
Ah, that was a mistake. Good that you found it!

PR #459, opened 2018-02-21
 Port the tutorial exercise generator to CakeML

### LINKED ISSUE #380: Implement make_ex in CakeML
Following #350 we now have a small SML program to generate the tutorial exercises, currently at `tutorial/solutions/make_ex.sml`. This issue is to re-implement this program in CakeML instead (including automatically generating the executable and using it in the relevant `Holmakefile`), in part to ensure our infrastructure is up to scratch.

The program might also be better in the `developers` directory, but that will depend on how the "build a CakeML program" aspect is handled.

michaelsproul:
Translated the `make_ex` utility for building tutorial exercises from their solutions to CakeML.

Closes #380


developers/Holmakefile
```diff
@@ -1,7 +1,30 @@
+INCLUDES = ../compiler/bootstrap/compilation/x64/64
```
xrchz: I worry that this is a circular dependency. Or at least an undesirable one, since `developers` may be required for some very basic stuff...

(The way around this would be to move these dependencies to a separate directory.)

tutorial/solutions/Holmakefile
```diff
@@ -2,18 +2,12 @@ INCLUDES = $(HOLDIR)/examples/balanced_bst ../../misc ../../translator ../../bas
...
+all: $(TARGETS)
 
 .PHONY: all
 
-exercises: $(SOLUTIONS) make_ex
-	./make_ex $(SOLUTIONS)
-
-make_ex: make_ex.sml
-	$(POLYC) make_ex.sml -o make_ex
+# To build the exercises from the solutions, run `Holmake exercises` from `../../developers`
```
xrchz: No, I think this should be an automatic target here. (Or at least an automatic target somewhere.)

michaelsproul: Oh yeah, an automatic target is a good idea. What if `make_ex` lived in a subdirectory of `developers`, with a dependency on the bootstrapped compiler. Could pave the way to writing a few more utilities in concrete syntax?

xrchz: Yep.

myreen:
@michaelsproul

We shouldn’t make the generation of the tutorial depend on the bootstrapped
compiler, because developers that modify tutorial/solutions hardly ever
wait for the bootstrapped compiler to run to see if the tutorial needs to
be updated from tutorial/solutions.

Two alternatives:

1. Require that some version of the bootstrapped compiler exists in the
user’s path.

2. Have the regression test workers commit updates to the tutorial when
they find that the script updates tutorial (after a full compiler bootstrap
has run).

michaelsproul:
Ok, I've pushed a new version that should address your comments:

* `make_ex` now lives in `developers/cml_utils`
* Building the CakeML utils requires a bootstrapped compiler, which will be sourced from `$PATH` if possible, or built from scratch if not
* The makefile in `tutorial/solutions` is now more or less as it was before, the main difference being that it now depends on `developers/cml_utils`, which means it will either build a bootstrapped compiler (probably not desired), or use one from `$PATH`

One caveat of how building works at the moment is that you need `basis_ffi.o` to compile new programs. My code just assumes that `basis_ffi.o` will be sitting next to `cake`, whether it is in `$PATH` or in the repo (which should usually be OK, as `basis_ffi.o` is built next to `cake` in the process of building `cake`).

michaelsproul:
Updated this morning with some more fixes, including a better solution for the basis library

myreen:
The regression test has revealed a problem with this PR. The problem is that `tutorial/solutions` seems to take a very long time to build:

https://cakeml.org/regression.cgi/job/237

My guess is that the problem stems from a dependency on the bootstrapped compiler: I suspect `Holmake` recursively start building the bootstrapped compiler during the regression test and all of this appears as just `Starting tutorial/solutions`.

Currently on master, `tutorial/solutions` is a nice early test case to see if everything works. Now `tutorial/solutions` has become the complete opposite.

@xrchz How did you envision the dependency on the CakeML compiler to be handled in regression tests (and in general)? I guess this applies in general case since users who want to build the tutorial solutions will just type `Holmake` in the directory and be surprised that the dependencies take a day to build and require a lot of RAM.

I am **closing** this issue for now to indicate that no further regression tests are to run on this until a solution to the problem is found. I hope a good solution can be found. 

Could the script be written so that it works in both Poly/ML and CakeML? My suggestion is that `Holmake` uses the CakeML compiler if it is available and Poly/ML otherwise.

xrchz:
I don't have a good solution in mind. Would require more thought... If there is a fallback to Poly/ML, ideally the regression would test both using Poly/ML and using CakeML. (Possibly at different points in the test sequence.)

myreen:
Yes, one could test whether the bootstrapped compiler produces the same output as the Poly/ML run of the tool earlier in the build sequence.

One solution could be to treat `cake` as a required program on each regression test worker. This limits how much the CakeML language can change because it's a hassle to install a new version of `cake` on all of the regression test servers.

michaelsproul:
Alternatively, we could give up writing `make_ex` in CakeML, and instead write some other test programs that are compiled and run after the bootstrapped compiler is built. We could include the expected output in the test cases.

The Rust compiler has an extensive set of tests like this, with different categories like `compile-fail`, `run-pass`, `run-fail`, that we could mimic: https://github.com/rust-lang/rust/tree/master/src/test

(although we probably don't need as many tests because we have proofs)

myreen:
Currently, we have one test after bootstrap: it compiles and runs a
hello-world example and checks that the output is the expected one. This
catches problems such as type inference failure in the basis library.

I know performance is hard to reliably test, but it would be interesting to
have performance tests, because we don’t prove anything about performance.
In particular, if the new compiler is suddenly producing 10 % slower code
that should probably be flagged.

Another alternative to the dependency on cake is that the build sequence
should download a given compiler from the webpage (instead of building it
via bootstrapping). This way each worker installs locally a specific CakeML
compiler and thus a manual update on all of the regression test workers
isn’t needed. I think I prefer this alternative.

xrchz:
I am against not writing `make_ex` in CakeML: the point of that exercise (which @michaelsproul has done) was to show up these kinds of problems with actually using CakeML, and resolve them. It's not just an example to test the CakeML compiler. It's something to force us to make tools that we can actually use in practice.

myreen:
I agree with @xrchz on this.

@xrchz What do you think of having the build download a CakeML 
compiler from the webpage? One could try to be specific e.g. have 
the commit hash be part of the URL of the zip file.

xrchz:
Downloading the built compiler sounds good. It could be the one at https://cakeml.org/cake-x64-64.tar.gz (which is the last successfully built master, as updated manually by me), or something more specific if you prefer.

PR #461, opened 2018-02-22
Remove num_exp from WordLang

ncough:
`num_exp` existing in WordLang to represent calculations
based only on constants, however, it was never used beyond
wrapping `num` values. This change removes `num_exp` and replaces
its uses with `num`.


xrchz:
Closed because there's more to come?

PR #463, opened 2018-02-24
Drule revised

myreen:
Switches CakeML proofs to use the `drule` version that exists in HOL4 (rather than CakeML's preamble)


mn200:
Latest local build succeeds in
* `compiler/proofs`
* `examples`

PR #470, opened 2018-03-02
better Isabelle support

larsrh:
This removes all traces of the `Isabelle_Support` theory. All constants are mapped to existing Isabelle concepts.

### DIFF
--- a/semantics/evaluate.lem
+++ b/semantics/evaluate.lem
@@ -148,6 +148,9 @@ evaluate_match st env v ((p,e)::pes) err_v  =
     end
   else (st, Rerr (Rabort Rtype_error))
 
+declare {isabelle} rename function evaluate = fun_evaluate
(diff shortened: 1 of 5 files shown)


semantics/evaluate.lem
```diff
@@ -148,6 +148,9 @@ evaluate_match st env v ((p,e)::pes) err_v  =
     end
   else (st, Rerr (Rabort Rtype_error))
 
+declare {isabelle} rename function evaluate = fun_evaluate
+declare {isabelle} rename function evaluate_match = fun_evaluate_match
```
larsrh: There's a bug in Lem that makes it produce invalid Isabelle output when constants are automatically renamed. This works around that bug. (I'll open an issue on the Lem repository soon, when I have minimized it accordingly.)

PR #471, opened 2018-03-03
Add a primitive ListAppend to all ILs up to wordLang

oskarabrahamsson:
This PR contains the following:

* Adds a primitive operation for the list-append operation to all languages up to wordLang. 
* The operation is compiled into efficient tail-recursive code in wordLang.
* Extends the bvi_tailrec pass to treat this operation.
* Also: various tweaks to bvi_tailrec which as a result should spend less time on functions it cannot optimise.


oskarabrahamsson:
I’ll merge and fix this locally.

oskarabrahamsson:
Sigh, it seems 75b9ed6 introduced some bogus line in `to_word32Script.sml`.

PR #472, opened 2018-03-08
Remove num_exp from WordLang

ncough:
`num_exp` existing in WordLang to represent calculations
based only on constants, however, it was never used beyond
wrapping `num` values. This change removes `num_exp` and replaces
its uses with `num`.


ncough:
Ran into some issues with this and got a little sidetracked. With the fix, the change passes translation on my machine.

PR #475, opened 2018-03-14
Add stage to download latest cake binary

ncough:
This changes adds a stage to the regression
worker to download the latest cake binary from
cakeml.org. This is then used by the sexpr bootstrap
and other testing stages.

Due to bootstrapping memory requirements, the downloaded
binary has its heap and stack increased to 2GB.


xrchz:
looks fine - will merge after the new stuff gets tested on the worker

ncough:
The worker successfully performed the `sexpr` bootstrap using the downloaded cake binary. @michaelsproul has access to the worker and has verified the bootstrapped binary is available.

PR #476, opened 2018-03-14
Annotate build-sequence with artefacts

ncough:
The regression worker has been recently extended to support
early upload artefacts based on annotations in the
build-sequence. This change makes use of these annotations to
upload results from verified and unverified bootstraps.

### DIFF
--- a/developers/build-sequence
+++ b/developers/build-sequence
@@ -82,11 +82,11 @@ compiler/parsing/tests
 compiler/bootstrap/translation
 
 # compiler sexpr bootstrap
-unverified/sexpr-bootstrap/x64/64
-unverified/sexpr-bootstrap/x64/32
+unverified/sexpr-bootstrap/x64/64:unverified/sexpr-bootstrap/x64/64/cake-unverified-x64-64.tar.gz
+unverified/sexpr-bootstrap/x64/32:unverified/sexpr-bootstrap/x64/32/cake-unverified-x64-32.tar.gz
 
 # compiler HOL bootstrap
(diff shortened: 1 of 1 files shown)


xrchz:
Maybe it would be nicer if the path of the upload were relative to the target directory. This is fine for now though (will see if it works then merge).

xrchz:
Am I right in thinking this should have uploaded artefacts already? (But apparently hasn't.) Maybe because the worker isn't up to date?

myreen:
I see that this recent build did upload artefacts:
https://cakeml.org/regression.cgi/job/262

xrchz:
@michaelsproul should the worker for job 261 have uploaded things early?

michaelsproul:
It should have, but it failed with this error:

```
Could not find cakeml-261/unverified/sexpr-bootstrap/x64/64/cake-unverified-x64-64.tar.gz to upload.
Could not find cakeml-261/unverified/sexpr-bootstrap/x64/32/cake-unverified-x64-32.tar.gz to upload.
```

It fails because the worker does the upload from the target directory, but the path we provided is a path relative to the repo root. So, the worker already expects a path relative to the target directory, as you suggested a few comments back!

If @ncough updates his PR to use just `:cake-unverified-x64-{64,32}.tar.gz` as the artefact path it should work fine

ncough:
Thanks @michaelsproul for the help. The PR has been updated to use relative paths and requires a worker with the changes from https://github.com/CakeML/regression/pull/5.

michaelsproul:
New build with updated worker running now

PR #477, opened 2018-03-26
[WIP] Pretty print library

### LINKED ISSUE #362: Write a pretty-printing library
This issue is to:
-  create a library for the CakeML basis above which one can write pretty-printing functions for datatypes, and,
- include some default pretty-printers for built-in datatypes (lists, options, unit/tuples, base types).

It should be possible to obtain a string from a pretty-printer, but this may not be the best underlying representation. It should also be possible to efficiently print the output of a pretty-printer (using `TextIO`).

zapashcanon:
Hi,

Fix #362.

This is a draft, just to be sure this is what was expected.

For every basic type `t` I wrote a `fromT` and a `printT` function.

`fromT` builds a `string app_list` from a `t`.
In case were `t` is polymorphic and depends on `'a`, it needs one more argument which should be a function `'a -> app_list`, this is the case in `fromList` as an example.

`printT` is basically the same, but prints directly instead of building a `string app_list`, which is more efficient.

Things that could be improved are the following:

- for Vector and Array, I can build a balanced `string app_list` instead (by doing some dichotomy algorithm), to reduce the possibility of a stack overflow when calling `print_app_list `. Same for the print function
- for List, I can write a CPS version of the print function, to avoid stack overflow, and maybe build a more balanced tree in the `fromList` function
- maybe writing a CPS version of the `print_app_list` function, to avoid all stack overflow, plus, showing that it is equivalent to the non-CPS one could be interesting :-)

### DIFF
--- a/basis/PrettyPrintProgScript.sml
+++ b/basis/PrettyPrintProgScript.sml
@@ -1,0 +1,212 @@
+open
+  preamble
+  ml_translatorLib
+  ml_translatorTheory
+  ml_progLib
+  basisFunctionsLib
+  StringProgTheory
+
+val _ = (
+  new_theory "PrettyPrintProg";
+  translation_extends "StringProg";
+  ml_prog_update (open_module "PrettyPrint")
+)
+
+(*val () = generate_sigs := true;*)
+
+val pp_fromString = process_topdecs`
+  fun fromString s =
+    List ["\"", s, "\""]
+`
+
+val pp_printString = process_topdecs`
+  fun printString s = (
+    TextIO.print "\"";
+    TextIO.print s;
+    TextIO.print "\""
+  )
+`
+
+val pp_fromChar = process_topdecs`
+  fun fromChar c =
+    List ["#\"", (Char.toString c), "\""]
+    (* TODO: check result of Char.toString for non-printable char, should be e.g. \3 and not something weird *)
+`
+
+val pp_printChar = process_topdecs`
+  fun printChar c = (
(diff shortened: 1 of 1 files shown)


myreen:
Thank you @zapashcanon for looking into this and asking whether you are going in the right direction. 

The high-level ideas sound good. Here are a few comments:
 - The functions you propose sound good in general, but I don't see the benefit of the more advanced stuff (i.e. CPS versions or balanced `app_list`s). If you can demonstrate a clear performance benefit, sure. If not, then keep the code simple and small.
 - I would like the module to be all about the `string app_list` type (which in HOL is `mlstring app_list`): all functions return this type and then there is a `print_app` function that given a `string app_list` prints it without producing the full string at any point.
 - Developments of this form should not use `process_topdecs`, instead they should define HOL functions from which CakeML code is generated using the translator, see `translate` in `ListProgScript.sml`; this way developments based only on HOL functions can use your functions. (Note that `print_app` would need to be developed with `process_topdecs` and be CF-verified)

Also as a separate comment: we tend to use open *pull requests* (PRs) to indicate that a branch is in the queue for merging into master. This PR is clearly not ready and as such shouldn't be an open PR. Discussions such as this should probably happen on the relevant issue, in this case #362, or on the slack channel.

myreen:
In the comments for #362, @xrchz pointed out that there already exists a `print_app_list` in [the basis library](https://github.com/CakeML/cakeml/blob/master/basis/basisProgScript.sml#L14).

PR #478, opened 2018-03-28
Allow users to prove necessary translator induction after translation

### LINKED ISSUE #442: Make translator state and prove tailored induction theorems
Currently, the translator uses induction theorems from `Define` to close the loop in the certificate theorems for recursive function.

Due to minor differences in what the translator expects and what `Define` produces, this part of the translator can be fragile. The current implementation contains several tactic proofs (one is run if the one before fails) to try to smooth over the minor differences.

This issue is about making the translator:

   1.  state what induction theorem it wants
   2.  prove the new induction theorem using the the one provided by `Define`
   3.  use new induction theorem to close the loop in translations of recursive function

A failure in 1. or 3. is bug in translator implementation. This should make the process a bit more user friendly in that a user can potentially manually prove the induction theorem when step 2 fails. By providing the exact induction theorem, step 2. is guaranteed to succeed.

myreen:
Closes #442


myreen:
The [regression test failed](https://cakeml.org/regression.cgi/job/275) because of an issue on master in `compiler/parsing/tests`. A local build of the bootstrap suggests that this PR is safe to merge.

PR #479, opened 2018-04-05
Fix sign error in reg_alloc

tanyongkiam:
dec_deg was increasing the degree rather than decreasing it


xrchz:
Happy to merge as soon as there's confirmation from anyone that this goes through the translator.

tanyongkiam:
reg_allocProg succeeded for me

PR #483, opened 2018-04-13
Make compilerTheory pick up Globals.HOLDIR

oskarabrahamsson:
Fixes the issue that some compiled binaries do not show the HOL commit by using `Globals.HOLDIR` instead of relying on the `HOLDIR` environment variable.


xrchz:
Can `mlstring_from_proc_from` raise exceptions?

oskarabrahamsson:
No, any exceptions should be caught:
https://github.com/CakeML/cakeml/blob/31c27782fd9fd62e0acfe9275cfe1587d12f71b1/misc/preamble.sml#L419

xrchz:
Ok, that is bad style (Interrupt should never be caught) but no fault of this PR.

PR #489, opened 2018-05-08
Replace sed and wget with alternatives

ncough:
Replace use of wget with curl and sed with a simple sml script to
minimise build dependencies.


ncough:
I made the requested modification. I agree, its a little messy but I am not sure what else to change. This actually ends up being faster than the original `sed` anyway (on my machine at least), so there doesn't seem to be a performance issue.

PR #492, opened 2018-06-12
Port Magnus' fixes to ml_translatorLib

tanyongkiam:
- This fixes some bugs with translation of mutually recursive functions

Edit: failure wasn't due to mutual recursion


xrchz:
Could we have a test for these bugs in ml translator test theory?

PR #496, opened 2018-07-02
Divsem

IlmariReissumies:
This pull requests extirpates the final event component from the ffi state, and instead stops evaluation whenever FFI divergence or FFI failure is encountered.

@SOwens: I'm tagging you as reviewer because I hope you might check if the lem changes make sense :)


IlmariReissumies:
It's unlikely that this PR has any impact on bootstrap since it doesn't touch the compiler, only semantics and proofs. And `x64BootstrapProof` is basically identical to the stuff in examples/proofs, which did go through.

With that said, I'm not in a rush; I'm happy to wait for a full bootstrap to go through if that's your preference.

xrchz:
My reasoning is that `x64BootstrapProof` depends on the compiler proofs, which have been changed.

PR #501, opened 2018-07-21
fix duplicate rule name

larsrh:
Should ideally be backported to the `version2` branch.


xrchz:
I'm not so keen on backporting -- there should be a 2.1 release in the next couple of months.

PR #502, opened 2018-07-25
Prove some stronger results about GC algorithm

myreen:


myreen:
I've checked that all of the compiler proofs succeed. This PR only touches the compiler proofs, so I'm merging this PR.

xrchz:
The compiler proofs also affect the bootstrap proofs (this is why #498 is not merged)

myreen:
#498 changed the basis which affects the translator and the CF proofs, so this is clearly affecting the bootstrap.

In contrast, the additions to the GC proofs in this PR are completely internal to the proofs of `data_to_word` and do not even affect the top-level proofs of the `data_to_word` compiler. I can’t see how these could break anything later. (I also didn’t introduce any new automatic rewrites.)

xrchz:
Ok that makes sense :)

PR #503, opened 2018-07-26
Fix unverified sexp printer for wildcard patterns

IlmariReissumies:
Previously it would print Pany as "Pany", but the sexp parser expects "(Pany)"

### DIFF
--- a/unverified/sexpr-bootstrap/astToSexprLib.sml
+++ b/unverified/sexpr-bootstrap/astToSexprLib.sml
@@ -93,6 +93,7 @@ fun op_to_exp arg =
 val cons = ``CONS : 'a -> 'a list -> 'a list``;
 val comma = ``$, : 'a -> 'b -> 'a # 'b``;
 val pvar = ``ast$Pvar``;
+val pany = ``ast$Pany``;
 val locs = ``Locs``;
 val nil_l = ``[] : 'a list``;
 val app = ``ast$App``;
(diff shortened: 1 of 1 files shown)


xrchz:
Why is there an unverified sexp printer at all? Is the verified one unusable in some way?

myreen:
The unverified sexp printer is used to print the result of the bootstrap translation (for prototyping purposes). The verified in-logic version simply doesn't scale sufficiently. 

I guess one could extract a reasonably efficient SML version from the in-logic version either via the CakeML translator or even emitML. However, one would still need to make the SML version interface with the bootstrap-generated AST which lives within the logic.  

(I think we've had this discussion a few times.)

xrchz:
How about creating another executable (alongside the compiler) by in-logic compilation, which is a simple executable that prints the result of `print_sexp bootstrap_ast` to `stdout`?

myreen:
It is possible to produce the CakeML AST for such a program within HOL at the end of the compiler bootstrap translation. However, to get that executable out you'd still have to run in-logic compilation on a large program. In this program, most of the code is building the AST value that then gets printed. This value-building code is still normal code as far as the compiler is concerned, but this approach might be worth a try!

IlmariReissumies:
Besides the bootstrap, I find it generally useful during development to have a quick-and-dirty way to get from a prog to an executable that I can test. The unverified sexp printer followed by  the bootstrapped compiler seems by far the fastest way to do this.

PR #505, opened 2018-08-01
DON'T MERGE: Dummy PR to test ulimit on regression servers

michaelsproul:


xrchz:
closing this - message me if you still need it; I expect we can find a better solution.

michaelsproul:
Ulimiting to 50GB mostly worked, although it was slower and the test never completed due to the server outage. I wouldn't mind trying again with a 90GB ulimit?

PR #508, opened 2018-08-06
additional length check in pmatch

larsrh:


larsrh:
I'm surprised that this was just such a tiny change after all. Thanks @myreen for fixing the proofs.

Issue #374, opened 2017-11-14
Translator generates false lookup_cons assumptions


tanyongkiam:
In ml_module_demoScript, run DISCH_ALL(hol2deep``Even m``); after closing the module gives:

val it =
   |- Eval env (Var (Short "m")) (NUM m) ⇒
   lookup_cons "Even" env =
   SOME (1,TypeId (Long "Even" (Short "even"))) ⇒
   Eval env (Con (SOME (Short "Even")) [Var (Short "m")])
     (EVEN_TYPE (Even m)):
   thm

The lookup_cons assumption will break, because the constructor is actually behind a module (lookup_cons only looks for Shorts)

This is due to an improper fix (by me) in 1e53ebd.

myreen:
There is a TODO regarding this at the [bottom of ml_translatorLib](https://github.com/CakeML/cakeml/blob/master/translator/ml_translatorLib.sml#L3930):

    TODO
    - ensure datatypes defined in modules can be used outside a module
      (the type thms need to be reproved)

oskarabrahamsson:
I have started to work on this issue on the [translator-module-ctors](https://github.com/CakeML/cakeml/tree/translator-module-ctors) branch.

oskarabrahamsson:
This was solved with #512.

oskarabrahamsson:
Yes, this particular issue is solved: `lookup_cons` previously only did lookups using `Short` names, so it was impossible to translate a function which pattern matched on constructors that were living inside a module. But something else is clearly still broken if `Dtype` declarations in the generated program are wrong.

PR #512, opened 2018-08-19
Make translator aware of datatype constructors that live inside modules

oskarabrahamsson:
This should hopefully fix issue #374.

The translator asks `ml_progLib` for the current module when a type is registered, so take care when calling `register_type` on types that would cause `register_type` to get called recursively, or those things might also end up as if they are living within your module.

### DIFF
--- a/basis/MapProgScript.sml
+++ b/basis/MapProgScript.sml
@@ -6,13 +6,11 @@ val _ = new_theory "MapProg"
 
 val _ = translation_extends "CompareProg";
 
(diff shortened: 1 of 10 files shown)


xrchz:
Nice work! Breaks the bootstrap translation currently. Reopen when fixed.

xrchz:
Now is the bootstrap translation diverging?

oskarabrahamsson:
Yes, there's likely a never-ending metis_tac in an EqualityType proof. I'll re-open this later.

Issue #260, opened 2017-04-04
CakeML should return non-zero on error


michael-roe:
On Unix-influenced operating systems, it is usual for programs to return a non-zero status on error. For example, compilers will return non-zero if an error was encountered during compilation.

At the moment, the bootstrapped CakeML compiler seems to return 0 whether or not compilation succeeded; it should return non-zero on failure.

xrchz:
Agreed. Probably @myreen knows best how to achieve this.

myreen:
One should probably have an FFI function called `exit_with_code`. The implementation on the C side is to call `exit` with the first element of the byte array that's given as argument. The semantics is that this is a non-returning FFI call, i.e. `Oracle_diverge `.

michaelsproul:
I'm going to try tackling this, as a way of becoming familiar with the FFI internals

michaelsproul:
Hi @myreen @xrchz! I tried to make some progress on this issue, but I'm stuck on the characteristic formula proofs. You can see my work-in-progress code here: https://github.com/CakeML/cakeml/compare/master...michaelsproul:issue260

It's not clear to me how divergence should work in a characteristic formula proof... both `POSTv` and `POSTe` seem wrong for a function that never returns. The first proof to break when I changed the definitions was `main_spec` in `compiler64ProgScript.sml`. From there I back-tracked to trying to prove something simple about the `exit_with_code` function itself.

I'd also like to understand exactly how the oracle (which I modified) interacts with the characteristic formulae, as I haven't been able to work that out (i.e. I haven't seen my change to the oracle effect anything). Pretty much any tips/explanations/pointers would be greatly appreciated! 🙂 

Thanks!

michaelsproul:
For now my code has the `exit_with_code` function in the `Comandline` module, but I plan to move it to `Runtime`

xrchz:
@michaelsproul how's it going?

michaelsproul:
I gave up on this for a while because of the lack of semantics for FFI divergence, but have been thinking about getting it going again now that Johannes has fixed that

xrchz:
@IlmariReissumies @michaelsproul could we get this finished soon do you think?

michaelsproul:
Work in progress branch is here: https://github.com/CakeML/cakeml/compare/master...michaelsproul:exit-code

Let me know if any of it looks unreasonable

michaelsproul:
[...]

PR #513, opened 2018-08-21
Support for FFI-divergence in CF

IlmariReissumies:
Among other things, I hope this would allow work on issue #260 to move forward.


myreen:
Currently, the exit-with-failure function called `Runtime.exit()` in the CakeML basis. To me, this choice of name doesn't signal that the error code will be `EXIT_FAILURE` on the C side in `basis_ffi.c`.

I think a name such as `Runtime.abort()` would be more intuitive. Alternatively, one could have the `Runtime.exit` function take an argument just like the `exit` function in C code.

xrchz:
+1 for `exit` that takes a status argument.

PR #514, opened 2018-08-21
Linear scan register allocator

### LINKED ISSUE #458: Linear scan register allocation
This issue (suggested by Magnus) is about writing a linear scan register allocator for CakeML.

This could be very useful when integrated with the CakeML compiler, e.g. when we get a read-eval-print loop, we would want to run a faster allocation algorithm.

This issue can have many scales: a BSc-level project might simply implement an unverified version (replacing the allocator in unverified/reg_alloc), while a MSc-level project could verify the algorithm, and integrate it with the word_alloc's correctness theorem.

myreen:
This PR adds a verified linear scan register allocator to the 
CakeML compiler. Users of the bootstrapped compiler can 
use the new register allocator using the `--regalloc=linear` 
command-line option.

This is work by @TWal. 

Closes #458


tanyongkiam:
May I suggest removing the --regalg=linear flag and just keep it to the same interface as the rest of the allocator, i.e. with --reg_alg=4 (which I think has the same effect?)

xrchz:
Seems to break backend proofs... Reopen when fixed.

myreen:
@tanyongkiam sure, my latest commit (b7646134b4ea6e43cfacab33af94840de7f2278f) removes the unnecessary command-line option.

PR #515, opened 2018-08-23
OpenTheory article reader

oskarabrahamsson:
Also changes the way whole_prog_spec is defined and used in basis_ffiLib/basis_ffiScript.


basis/basis_ffiScript.sml
```diff
@@ -288,7 +288,7 @@ val iobuff_loc_num =
   TextIOProgTheory.iobuff_loc_def
   |> concl |> rhs |> rand;
 
-val IOFS_precond = Q.prove(
+val IOFS_precond = Q.store_thm ("STDIO_precond",
```
xrchz: This looks like a mistake...

candle/standard/monadic/holKernelProofScript.sml
```diff
@@ -31,6 +31,20 @@ val REPLICATE_11 = Q.prove(
...
 
+(* ------------------------------------------------------------------------- *)
+(* case_eq theorems                                                          *)
+(* ------------------------------------------------------------------------- *)
+
+fun get_thms ty =
+  { case_def = TypeBase.case_def_of ty
+  , nchotomy = TypeBase.nchotomy_of ty }
+
+val case_eq_thms =
```
xrchz: For future reference, you might also like to use `CaseEq`

candle/standard/monadic/holKernelProofScript.sml
```diff
@@ -386,6 +427,45 @@ val term_type = Q.prove(
   imp_res_tac CONTEXT_std_sig >>
   fs[TYPE_def,type_ok_def,is_std_sig_def])
 
+val type_of_has_type = Q.store_thm("type_of_has_type",
```
xrchz: I'm suspicious of this needing an inductive proof... isn't something like it already proved?

candle/standard/opentheory/prettyScript.sml
```diff
@@ -0,0 +1,491 @@
+open preamble holSyntaxTheory mlstringTheory
+
+val _ = new_theory "pretty";
```
xrchz: This theory should not be here. I think it should be in the basis library.

xrchz: At least, the part that's not specific to HOL terms. (Although @myreen also wants the Candle kernel to be in the basis library..)

PR #517, opened 2018-08-27
Check that prog_def is a nil-terminated list before compiling

oskarabrahamsson:


compiler/compilationLib.sml
```diff
@@ -892,6 +892,8 @@ val intermediate_prog_prefix = ref ""
 
 fun compile backend_config_def cbv_to_bytes heap_size stack_size name prog_def =
   let
+    val _  = can listSyntax.dest_list (rhs (concl prog_def)) orelse
```
xrchz: I think this is `listSyntax.is_list` for future reference

PR #519, opened 2018-09-01
Fix compilationLib

oskarabrahamsson:


xrchz:
This undoes the check for a concrete list - I think that was a good check to have.

oskarabrahamsson:
Agreed, I'll put it back in. But I'd like it if we merged this PR ahead of that, and working master branch again.

PR #522, opened 2018-09-11
Prepare to branch off after stackLang

lorenzleutgeb:
TL;DR:
 * New notion and corresponding types: `preset` derived from `config`.
 * Refactoring: `backend_conf.lab_conf.asm_conf` is now `backend_conf.asm_conf`
 * Refactoring: `backend$compile` does not return `lab_to_target$config` anymore, but instead `string list` (FFI names).
 * New type: `backend$late_config = ToTarget ...` which will later also have `| ToWasm ...` to control branching after `stackLang`.
 * New command line switch: `--wasm=` (boolean, for 64bit archs only) which currently is a no-operation. (Note that this PR is not against `master` but against `wasm`.)

---

This PR aims to prepare the compiler for integration of the WebAssembly target, which will introduce a branch/fork: Instead of compiling `stack → lab → target → export`  it must accomodate `stack → wasm → export` as an alternative.

In the record `backend$config`, replace the fields `stack_conf` and `lab_conf` with a field `late_conf: late_config`. A late config is meant to configure what the "late stages" of the backend, i.e. after `stackLang` should do. Currently, there is only one constructor `ToTarget stack_to_lab$config lab_to_target$config` which is meant to behave the same as the "old" variant. However, I plan to expand this by providing a second constructor `ToWasm stack_to_wasm$config`.

With this change, I am introducing a mismatch between the "configuration" that target architectures define and the actual backend configuration. This is resolved by introducing "presets": A preset is a meant to as a blueprint for a configuration and defined by the architecture (`{x64,arm8,...}_preset`, previously `*_config`). A preset is then converted into a configuration at runtime, and further extended by merging with commandline prameters as before.

While theoretically the future wasm target would not require the architecture-specific details contained in those presets, this approach allows to generate wasm code that might actually be nicer to JIT for a particular architecture. Also, it allows to define a wasm preset, possibly synthesized from experiments regarding performance in the browser to in turn influence other translations.

While implementing these changes I noticed a hack: Early translations such as `word_to_stack` and `data_to_word` use the configuration for `lab_to_target` to access its `asm_config`. This does not fit well with the new-style configs, since there might not even be a `lab_to_target` translation to be planned, thus leaving `word_to_stack`/`data_to_word` with a dangling assumption. This is mitigated by promoting `backend_conf.lab_conf.asm_conf` to `backend_conf.asm_conf`.

Similarily, `backend$compile` previously returned `lab_to_target$conf`, which is something that the `stack_to_wasm` translation cannot provide. I saw that in `compileScript` only the field `ffi_names` of the returned config was used. That's why I changed `lab_to_target$compile` to return the FFI names only. In case exporters at some point need more information, this can be solved by introducing a `export_config` record, that the ultimate `compile` functions must return.

*I still need help* fixing stuff: Since I changed the definition of `lab_to_target$config`, also `lab_to_target$compile` and `lab_to_targetProof` need to be adjusted. Perhaps @xrchz or @tanyongkiam could take over this part, or at least help me do it?

I know that there are other theories dependent on those `*config_Theory`s that I changed, but I would like to see some sort of approval of my approach before I go ahead and refactor them.

### DIFF
--- a/compiler/backend/backendScript.sml
+++ b/compiler/backend/backendScript.sml
@@ -15,15 +15,37 @@ open jsonLangTheory presLangTheory
 
 val _ = new_theory"backend";
 
-val _ = Datatype`config =
-  <| source_conf : source_to_flat$config
-   ; clos_conf : clos_to_bvl$config
-   ; bvl_conf : bvl_to_bvi$config
-   ; data_conf : data_to_word$config
+(* Target architectures provide config presets,
+ * which are turned into configs. *)
+val _ = Datatype`config_preset =
+  <| source_conf       : source_to_flat$config
+   ; clos_conf         : clos_to_bvl$config
+   ; bvl_conf          : bvl_to_bvi$config
+   ; data_conf         : data_to_word$config
    ; word_to_word_conf : word_to_word$config
-   ; word_conf : 'a word_to_stack$config
-   ; stack_conf : stack_to_lab$config
-   ; lab_conf : 'a lab_to_target$config
+   ; word_conf         : 'a word_to_stack$config
+   ; stack_conf        : stack_to_lab$config
+   ; lab_conf          : lab_to_target$config
+   ; asm_conf          : 'a asm$asm_config
+   |>`;
+
+(* Configures the later stages of the backend, i.e.
+ * where to go after stackLang. *)
+val _ = Datatype`late_config =
+  | ToTarget stack_to_lab$config lab_to_target$config
+  (* TODO: The following is in preparation of the wasm target: *)
+  (*| ToWasm   stack_to_wasm$config *)`
+
+(* Actual configuration, usually derived from a config_preset. *)
+val _ = Datatype`config =
+  <| source_conf       : source_to_flat$config
+   ; clos_conf         : clos_to_bvl$config
+   ; bvl_conf          : bvl_to_bvi$config
+   ; data_conf         : data_to_word$config
+   ; word_to_word_conf : word_to_word$config
+   ; word_conf         : 'a word_to_stack$config
+   ; late_conf         : late_config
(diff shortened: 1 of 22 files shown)


compiler/backend/backendScript.sml
```diff
@@ -108,31 +141,39 @@ val to_data_def = Define`
...
-  (c,p:'a prog)`;
+  case c.late_conf of
+  | ToTarget stack_to_lab_conf lab_conf =>
+    let (c,p) = to_stack c p in
+    let p = stack_to_lab$compile
+      stack_to_lab_conf c.data_conf (2 * max_heap_limit (:'a) c.data_conf - 1)
+      (c.asm_conf.reg_count - (LENGTH c.asm_conf.avoid_regs +3))
+      (c.asm_conf.addr_offset) p in
+    SOME (c,p:'a prog)
+  | _ => NONE`;
```
lorenzleutgeb: In case there is no configuration for `stack→ lab`, this function may now return `NONE`.

compiler/backend/backendScript.sml
```diff
@@ -108,31 +141,39 @@ val to_data_def = Define`
...
 
 val to_target_def = Define`
   to_target c p =
-  let (c,p) = to_lab c p in
-    attach_bitmaps c.word_conf.bitmaps
-      (lab_to_target$compile c.lab_conf p)`;
+  case c.late_conf of
+  | ToTarget stack_to_lab_conf lab_conf =>
+    case to_lab c p of
+    | SOME (c,p) =>
```
lorenzleutgeb: We not only need to check whether there is a config for `lab_to_target` but also that the previous step was successful.

xrchz:
Is this pull request in fact a request to merge these changes now? I don't think so. But reopen if so.

PR #523, opened 2018-09-12
Mention unit of measure for heap/stack sizes

lorenzleutgeb:
I needed to know whether the numbers are actually bytes or "something bigger". I figured that it must be mebibytes judging from [`x64_exportTheory`](https://github.com/CakeML/cakeml/blob/1bb8663a0aeb377eb22aeb600de4bc536e647744/compiler/backend/x64/export_x64Script.sml#L67) and its use of [`exportTheory`](https://github.com/CakeML/cakeml/blob/1bb8663a0aeb377eb22aeb600de4bc536e647744/compiler/backend/exportScript.sml#L29-L30).
For people more involved with CakeML, this is probably trivial. But when trying to find out for the first time, it would be quite helpful to have this information more prominent.

### DIFF
--- a/compiler/bootstrap/compilation/x64/32/Makefile
+++ b/compiler/bootstrap/compilation/x64/32/Makefile
@@ -3,7 +3,8 @@
 
 # To set the stack and heap directives available to your own
 # CakeML programs, use the --stack_size and --heap_size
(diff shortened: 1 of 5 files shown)


xrchz:
I'm merging this without a regression test since the changes are entirely to comments.

PR #524, opened 2018-09-12
Use Git's -C option

lorenzleutgeb:
See https://git-scm.com/docs/git#git--Cltpathgt

~~Note that `mlstring_from_proc_from` is dead code then.~~


xrchz:
Can you also delete the dead code?

lorenzleutgeb:
Of course.

PR #526, opened 2018-09-14
Resuscitate presLang as to_displayLang

talsewell:
RFC at this stage. This is a first step towards resurrecting the
explorer (or some of it): rebuild the conversions from the various
ILs to the printable displayLang form.

The theory runs, but I don't know how to test whether there's
any risk of breaking anything else (e.g. translation).

### DIFF
--- a/compiler/backend/arm6/arm6_configScript.sml
+++ b/compiler/backend/arm6/arm6_configScript.sml
@@ -44,7 +44,8 @@ val arm6_backend_config_def = Define`
                word_to_word_conf:=^(word_to_word_conf);
                word_conf:=^(arm6_word_conf);
                stack_conf:=^(arm6_stack_conf);
-               lab_conf:=^(arm6_lab_conf)
+               lab_conf:=^(arm6_lab_conf);
+               tap_conf:=default_tap_config
                |>`;
 
 val _ = export_theory();
--- a/compiler/backend/arm8/arm8_configScript.sml
+++ b/compiler/backend/arm8/arm8_configScript.sml
@@ -41,7 +41,8 @@ val arm8_backend_config_def = Define`
                word_to_word_conf:=^(word_to_word_conf);
                word_conf:=^(arm8_word_conf);
                stack_conf:=^(arm8_stack_conf);
-               lab_conf:=^(arm8_lab_conf)
+               lab_conf:=^(arm8_lab_conf);
+               tap_conf:=default_tap_config
                |>`;
 
 val _ = export_theory();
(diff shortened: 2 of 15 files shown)


talsewell:
Curiously, git hasn't figured out the files are related. I wonder if I should try
to adjust the edit so it does, so as to preserve the history to the real authors.

xrchz:
The regression test indicates that it did break something else in the compiler proofs.

xrchz:
We usually use open PRs for things that are ready to merge as soon as the regression test succeeds -- I don't think this is currently in that state, is it? It's possible this policy should change since PRs are nice for discussion. (The regression test server would need to be configured to pay more attention to labels or something other than just the PR open/closed state.)

xrchz:
> We usually use open PRs for things that are ready to merge as soon as the regression test succeeds -- I don't think this is currently in that state, is it? It's possible this policy should change since PRs are nice for discussion. (The regression test server would need to be configured to pay more attention to labels or something other than just the PR open/closed state.)

Note that this https://platform.github.community/t/negate-a-filter-find-prs-without-a-certain-label/6967 makes the above non-trivial. I'd be happy for someone else to think about a good way to change the server, though.

IlmariReissumies:
Or we could say that discussions on things that are not ready for prime time should go not on pull requests but on the issue that's being worked on.

xrchz:
> Or we could say that discussions on things that are not ready for prime time should go not on pull requests but on the issue that's being worked on.

Oh yes, that's a good norm too -- I had forgotten about issue discussions :)

lorenzleutgeb:
IMO Issues are fine for keeping track and discussing ideas and matters that are bigger/more abstract than code (or might not even result in code being produced). Also one issue might be associated with multiple PRs and have a longer lifetime. As soon as there actually is code to be discussed, a PR offers a better set of features (reviews that can be requested and then result in rejection/approval, comments on specific lines that expand into threads, integration with other checking services like Travis CI, ...). By closing issues for the sake of controlling the regression server, this "workflow" is disturbed a lot.

talsewell:
@xrchz OK, I'm happy to do things differently. In my first week I didn't have a clue how to run the regression test, or where to run it without breaking other things, so it was convenient to have the PR worker do it for me. Also, I would actually like to merge this? Should I go back and edit my leading comment? I'd clean the history first, and address any actual requests/complaints/preferences.

xrchz:
No need to edit the comments/discussion, but to get this merged it'll need to pass the regression test at least.

talsewell:
Well, yeah. I kind of hoped the last run would succeed. The partiality of translation is the problem here, and I have no experience in fixing it, but we'll get there.

talsewell:
More work to be done. I'm going to close this here to stop the regression worker wasting time, and reopen in the standard CakeML workflow.

PR #528, opened 2018-09-22
Fix a proof and add some timing.

talsewell:
Putting two things in the one PR to save regression runs, sorry.

I think this will fix at least one proof failure in master, possibly completing the run.

This also adds a translator timing logging feature that might answer some questions about the bootstrap translation timing.


xrchz:
Seems to be diverging. Can we check that?

xrchz:
How sneaky to combine two sets of changes :) Possibly the artefacts file doesn't do what you think by the way.

talsewell:
Indeed, it was sneaky. I was hoping to build some timing information over the weekend and read it on Monday. That didn't exactly work out, but I'm reading it now.

PR #534, opened 2018-10-08
Split data_to_word's assign_def into a function for each op

myreen:
This makes running `Holmake` on `data_to_wordScript.sml` much faster 
and the `assign` function is no longer as intimidating.


myreen:
The regression test revealed a problem. I'm temporarily closing this PR.

PR #537, opened 2018-10-09
Use modules in bootstrap translation.

talsewell:
If this succeeds and speeds things up, I'd like to merge it.

It groups the bootstrap translation elements into modules.

(It also removes a duplicate end_theory in reg_allocProgScript).

This should speed up bootstrap translation a bit, since the
number of things in the top-level namespace has an impact on
translator performance. Grouping many things in modules keeps
them out of the top-level namespace.


xrchz:
The speedup seems modest - maybe reduced the time by 1/8th?

talsewell:
Hmm, yes, the timing implications weren't what was expected. I'll have to investigate further, which unfortunately is a slow process.

PR #551, opened 2018-11-08
Translator performance changes

talsewell:
This is a union of two changes to the translator.

The first is to introduce yet another tree representation in HOL4. This one represents partial function trees as assoc-lists using ALOOKUP. These are given a tree structure where the assoc-list is sorted and the "node" operator is actually the append operator. The point is that balancing is actually a conv, and that decisions about sorting and balancing are made in ML rather than within the logic.

(EDITED) And the point of introducing the above type is to represent the lookups
(nsLookup auto_env1) using these alists. The representations are quite a bit faster to
build and look up than the previous solutions, which depended on the EVAL set caching a lot
of stuff for performance.

There's also a tweak in here to reduce the number of times that the translator code runs conversions that examine the program (SNOC list) within the translator state. This part isn't abbreviated, so simple adjustments to the convs and lookups to avoid looking at it can save a lot of CPU time.

### DIFF
--- a/characteristic/cfMainScript.sml
+++ b/characteristic/cfMainScript.sml
@@ -36,8 +36,7 @@ val call_main_thm1 = Q.store_thm("call_main_thm1",
   \\ fs [terminationTheory.evaluate_def,PULL_EXISTS,pair_case_eq,
          result_case_eq,do_con_check_def,build_conv_def,bool_case_eq,
          ml_progTheory.lookup_var_def,option_case_eq,match_result_case_eq,
-         EVAL ``nsLookup (merge_env env2 env1).v (Short fname)``,app_def,
-         app_basic_def]
+         ml_progTheory.nsLookup_merge_env,app_def,app_basic_def]
   \\ first_x_assum drule \\ fs [] \\ strip_tac \\ fs []
   \\ fs [cfHeapsBaseTheory.POSTv_def]
   \\ Cases_on `r` \\ fs [cond_STAR] \\ fs [cond_def]
--- a/characteristic/cfTacticsLib.sml
+++ b/characteristic/cfTacticsLib.sml
@@ -95,6 +95,8 @@ val reducible_pats = [
   ``do_con_check _ _ _``,
   ``build_conv _ _ _``,
   ``nsLookup _ _``,
+  ``nsLookup_Short _ _``,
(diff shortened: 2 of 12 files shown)


mn200:
The alist_tree stuff looks cool.  Can it be something for core HOL?  And how does it relate to enumfset, if at all?  (See the "Tree-Structured Finite Sets and Finite Maps" section of the DESCRIPTION manual.)

Your `Lib.sml` code needs to either not use the parser at all, or fix a parsing context so that its invocations of the parser are safe.  For example, put the following at the head of the file:
```
structure Parse = struct
  open Parse
  val (Type,Term) = parse_from_grammars boolTheory.bool_grammars
end
open Parse
```

xrchz:
It's better not to use the parser: that's one reason why we have Syntax libraries.

xrchz:
Closing does not kill the discussion. My current approach is to close it and add the `test failing` label, so I know which closed PRs are actually just suspended. I'll do that...

talsewell:
OK, I think the additional commit will fix the regression failure, so I'm reopening to retest.

As @mn200  pointed out, I still have an invocation of the parser in alist_treeLib to remove
before this can be merged.

talsewell:
Well, it looks like the test will now succeed ( https://cakeml.org/regression.cgi/job/556 ).

It did save a fair bit of time, but not much space.

xrchz:
The time savings are impressive! I will actually wait for the bootstrap (at least 64-bit) to finish before merging, but looking forward to having this.

talsewell:
No, I didn't check space usage independently.

I suspect that the regression reports of both time and space are a bit approximate (for instance, the oven1 workers probably speed up a bit when they're running alone). I don't have any clever ideas what to do about that.

xrchz:
oven1 looks to be stuck on this... I don't think it's due to these changes though...

talsewell:
Yeah, I made a mistake with oven1 - I figured out that interrupting curl can cause the worker to notice new jobs. Unfortunately I made it start up two jobs more or less simultaneously, which didn't work out.

There should be a solid performance improvement. @myreen - I think the 8.5 hours was actually a blip and a more common translation time is around 6 hours. The slowest theories speed up quite a bit, but smaller theories don't see a lot of improvement.

PR #553, opened 2018-11-14
flat_elim/word_elim cleanup

hrutvik:
#### Tidying dead code elimination - in `compiler/backend`
 * Improve the efficiency of the flatLang implementation - modify `flat_elim$close_spt` to more effectively compute reachability in a `num_set num_map`, and prove accompanying theorems.
 * Refactor - move all theorems to the relevant proof files, leaving `flat_elimScript.sml` and `word_elimScript.sml` as purely implementation and removing `reachabilityScript.sml`.
 * Some stylistic changes - use `snake_case` instead of `camelCase`, remove major uses of `>|` or `THENL`, keep under 80 characters per line, add comments at beginnings of files.

**TODO:** many lemmas at the top of `flat_elimProofScript.sml` could be moved to `sptreeScript.sml` - these are marked.

#### Examples - in `examples`
 * Create a double example in `doubleProgScript.sml`, with three implementations of a doubling function.
 * Create an array search example in `array_searchProgScript.sml`, with two implementations of array search (linear and binary).


xrchz:
The description, and a quick skim of the changes, look great to me. Thanks for this!

myreen:
Based on the regression test, there seems to be something broken in `compiler/backend/backendProofScript.sml`.

xrchz:
reopen when fixed

hrutvik:
The `compile_correct` theorem in `compiler/backend/backendProofScript.sml` was broken due to the renaming of definitions in `compiler/backend/flat_elimScript.sml` - this should be fixed now.

xrchz:
I suspect the latest failure is due to HOL changes introduced by @mn200

myreen:
My guess is that `spt_fold` is missing from `backendComputeLib`. I'm testing this locally at present.

PR #556, opened 2018-11-19
Rename "backend_correct" to "encoder_correct"

### LINKED ISSUE #539: Rename backend_correct to encoder_correct
It is confusing that the main proofs under `compiler/encoders` talk about `backend_correct` when there is also a `compiler/backend`. This issue is about taking a step towards consistency in terminology by calling the constant `encoder_correct` instead.

oskarabrahamsson:
Should fix #539.


xrchz:
I'm merging this early in the interest of moving things along -- I think the regression test has got past the relevant bits.

PR #558, opened 2018-11-21
Add --types option to bootstrapped binary

myreen:
The new `--types` option runs parsing and type inference, and then prints the types of all the visible top-level bindings. A side effect of this is that the empty program prints the types of all the user-visible bindings that come from the basis library. 

This closes issue #344.


myreen:
I'm closing this one until I've fixed the problem found by the regression test.

xrchz:
It looks like now you don't want to repoen this?

myreen:
Oops, I forgot that I had made this PR. There is a new one (#561) that is open.

Issue #360, opened 2017-11-01
Use theorem sets for CF specs


xrchz:
In `cfTacticsLib`, this function uses `IRULE_CANON` which is rather expensive. I don't believe this is necessary, and using a more efficient method could speed up `xapp`, `xlet_auto`, etc.

Vaguely related to #345

mn200:
It's possible that 

1. the excessive calculations involving free variables in `IRULE_CANON` could be simplified while yielding the same result; or
2. `IRULE_CANON`'s specification could be simplified to not require all those calculations without causing any real usability issues.

xrchz:
Improvements to `IRULE_CANON` sound worthwhile even if we don't end up using it in `is_cf_spec_for`. I think `is_cf_spec_for` should just search for any theorems that include `app`, and the functions that call it should backtrack over them.

Armael:
Indeed. I see two ways of fixing this:
- either reimplement `concl_tm` in a more efficient way; keeping the implicit specification that it returns the subterm that will be matched against the goal when using `irule` (in `xapp_xapply_no_simpl`)
- still use `IRULE_CANON` (since it will be needed when using `irule` later in `xapp_xapply_no_simpl`), but remember it, and thread the result through the code, and in `xapp_xapply_no_simpl` use a version of `irule` that takes the result of `IRULE_CANON` instead of re-computing it...

Armael:
@xrchz searching for any theorem that include `app` would return all the registered specs, and then try `irule` on each of them? This sounds costly.

xrchz:
No I mean search for `app` applied to the correct function. In other words, very similar to what is currently done but without the `IRULE_CANON`.

I think your first option is better than your second. But I'd rather see data on what works best.

Armael:
An other way would be to have a separate mechanism to store CF specifications (other than just `store_thm`), that would maintain a map `term` -> `spec`, which would avoid scanning the whole set of registered theorems at each call to `xapp`.

xrchz:
Yes that is clearly a desirable thing to do. It's easy to create a labelled set of theorems - this is already done for stateful simplifier (`[simp]` attribute) and congruence rule (`[defncong]` attribute) for example. I'd say this issue can be closed if we (a) create a theorem set for CF specs and (b) either remove `IRULE_CANON` from `is_cf_spec_for` or improve the performance of `IRULE_CANON`.

mn200:
[...]

PR #564, opened 2018-11-25
Equality type

talsewell:
Replaces the EqualityType prover with one that can handle some more complex inductions.

I can then delete all the manual EqualityType proofs from parserProg.

I think I can eventually delete all the manual EqualityType proofs, however I haven't made those edits because I think it would lead to merge conflicts with the cleanup that @oskarabrahamsson is doing.
  - I expect this PR will be edited to fit with that work before it is merged.
  - I still want to test & discuss the PR as-is.

There's one issue in the code I'd like some feedback on. I'm using dest_thy_const/DB.fetch in a clunky way to guess the definitions of the type invariant functions (LIST_TYPE etc) and to fetch a theorem about the num2x operator defined by EnumType. There's a pattern where the constant is always defined together with some key theorem, and you can guess the name of the key theorem. How ugly is that? Is it OK to keep doing it? I could ml_translatorLib to track both the type invariant constant names and their definitions, but I can't do the equivalent for EnumType/TypeBase.

### DIFF
--- a/compiler/bootstrap/translation/parserProgScript.sml
+++ b/compiler/bootstrap/translation/parserProgScript.sml
@@ -57,123 +57,19 @@ val _ = (find_def_for_const := def_of_const);
 val res = register_type``:(token,MMLnonT,locs) parsetree``;
 val res = register_type``:MMLnonT``;
 
-local
-  val ths = ml_translatorLib.eq_lemmas();
-in
-  fun find_equality_type_thm tm =
-    first (can (C match_term tm) o rand o snd o strip_imp o concl) ths
-end
+fun EqualityType_conseq_conv dir tm = let
+    val cc = ConseqConv.CONSEQ_REWRITE_CONV (eq_lemmas (), [], [])
+  in cc dir tm end
 
-val EqualityType_CHAR = find_equality_type_thm``CHAR``
-val EqualityType_INT = find_equality_type_thm``INT``
-val EqualityType_NUM = find_equality_type_thm``NUM``
-val EqualityType_LOCN = find_equality_type_thm ``LOCATION_LOCN_TYPE``
-val EqualityType_LOCS = find_equality_type_thm ``LOCATION_LOCS_TYPE``
+val EqualityType_tac = ConseqConv.CONSEQ_CONV_TAC EqualityType_conseq_conv
+    \\ simp_tac bool_ss [];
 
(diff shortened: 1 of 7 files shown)


mn200:
It seems like you have two issues here.

1. You have to guess at the bijection-with-num function names.  This is clearly sub-par.  I would have no problem if you just injected a new entry point into the TypeBase to store these.  Indeed, you could define them for all types (not just enumerated ones).  The function would not be a bijection for non-enums, instead it would just map all values of a given constructor to the same number.  This then also gives a method of avoiding quadratically-large constructor-distinctness theorems for all types, not just enums.

2. You need to be able to retrieve type invariant functions in a nice way.  You could use `ThmSetData` for this (in `src/1`), perhaps with a storage attribute as well.  You might also use the more primitive `ThyDataSexp` (from `src/postkernel`).  The latter is used to store the information that is then loaded into the `TypeBase` for example, which might be a guide as to how to use it. The aim with this stuff is to replicate the way Isabelle allows arbitrary data to be added to theories and queried in a clean way.  The HOL4 story is complicated by the desire to pickle/unpickle all this stuff to disk.

talsewell:
@mn200 indeed. Yes, a constructor-numbering function would be useful in a number of other (somewhat specialised) contexts, and could be officially added to TypeBase. I guess I might get around to that eventually?

The situation in (2) is actually slightly simpler, the translator state already tracks the type invariant functions to build terms on demand. I could adjust it to also track the relevant theorems. That's a bit easier than (1), I'm just not sure how valuable it is.

xrchz:
Point 2 sounds possibly related to #360

mn200:
@xrchz Agreed.  The `ThyDataSexp` stuff could also be used to replace the existing way that theorems are stashed in `packLib`.  (I'm sure the core functionality is the same so that the only advantage would be that the sexp-stored theorems are stored separately and so wouldn't be found with queries like `DB.match`.)

myreen:
I'm temporarily closing this so that it doesn't take up space on the regression machines.

xrchz:
Do you want this merged now, or do you want to keep working on further improvements mentioned in the original description?

talsewell:
I'd be happy to merge it now. I have other things to work on before I get around to refactoring the way ml_translatorLib stores state, and I don't think I'll get to look at TypeBase for a long time.

Issue #547, opened 2018-10-23
Linting


xrchz:
We would like to automatically check various stylistic conventions in the CakeML code base (i.e., to run a linter as part of the regression test). This issue is to implement the first version of the linter, probably as an extension to the existing README checker and generator, and to incorporate it in the regression test.

For the first version, the linter should check the following uncontroversial conventions:
- max line width of 80 columns
- no trailing whitespace
- no tabs
- (inherit `readme-gen`'s checks)
- optional: check `store_thm` (etc.) has a matching name and bound variable

As is common with linters, it should be possible to circumvent the automatic checks with an explicit comment.

(Note: other, more controversial, ideas for linting conventions are being collected on the Candle backlog.)

mn200:
The first three of these are now checked by the (misleadingly named) `unicode-grep` tool in the HOL distribution.

xrchz:
Would using `unicode-grep` for this be a reasonable gateway to linting more complex and controversial CakeML conventions? How much are you willing to add to that tool? We could use it for CakeML by making the regression test worker call it, but if we're also going to have our own linting tool, I don't mind duplicating the simple syntactic checks above.

mn200:
Yes, I think using it would be reasonable, and I'm certainly willing to add further to it.  I will try to add a `--lint` option to Holmake to enable the specification of a tool that will be run before anything else happens in a directory.  This could then be passed `unicode-grep` as an argument, or anything else more CakeML specific.  Of course, this option could be made "easy to access" by putting it into a `CLINE_OPTIONS` spec inside a Holmakefile.

myreen:
Related to the "optional: check `store_thm`" point, @mn200 and I are considering changes to HOL that would make HOL treat
```
Theorem foo[simp] goal tactic
```
as if it was
```
val foo = Q.store_thm("foo[simp]",goal,tactic)
```
As of a few minutes ago, [HOL master](https://github.com/HOL-Theorem-Prover/HOL/commit/7f7650b1f7d9fbc79f55646dabcf225b5cf0fff4) has this behaviour.

One could eventually move towards a state of the codebase where `Theorem` is used instead of `store_thm`. The linter could eventually enforce this, if it is the direction we want to go in.

PR #566, opened 2018-11-28
Automatically generate README.md files

### LINKED ISSUE #71: auto-generated readmes
generating the readme content from comments in the files in the directory could be one way to avoid it going too stale.

myreen:
The `readme_gen` script runs locally every time `Holmake` is run. It is run globally when the `developers` directory is built. The `developers` directory is one of the first directories to be built during a regression test.

Closes #71


myreen:
Closes #547

But note that instead of checking "max line width of 80 columns", `readme_gen` currently checks that no line exceeds 200 columns. With some additional work, this number can be reduced to 80.

Also note that these commits do not address "optional: check `store_thm`", since @IlmariReissumies is working on that separately.

PR #569, opened 2018-11-29
Replace store_thm and Q.store_thm with Theorem

IlmariReissumies:


IlmariReissumies:
Related to issue #547.

IlmariReissumies:
As per discussion on slack, we'll do this change on the `cleanup` branch instead

PR #574, opened 2018-12-06
Adding timestamps to values in dataLang

agomezl:
This PR introduces the timestamps for `Block` values  in `dataLang` along with some quality of life changes:

* `dataLang` no longer uses `bvlSem$v` values, but rather its own slightly tweaked version (with timestamps)

*  `dataSem$do_app` is no longer defined in terms of `bvlSem$do_app` and `bviSem$do_app`

* `data_to_word` theories no longer have a dependency in `bvi` or `bvl` theories

* `dataSem$state` now has a `tstamps : num option` field, to keep track of the current timestamp

* The semantics of `dataLang` where just slightly changed and should completely ignore the use of timestamps if `tstamps = None`


agomezl:
Closing while issues are being dealt with

PR #576, opened 2018-12-06
Move parsingComputeLib into parsing/proofs

### LINKED ISSUE #575: parsingComputeLib doesn't build
`parsingComputeLib` doesn't build under some very specific conditions anymore, Holmake in `compiler/parsing` fails with:
`Don't know how to build necessary target(s): /home/user/cakeml/semantics/semanticsComputeLib.ui`
This happened, because `semanticsComputeLib` was moved into `semantics/proofs`.
This can be potentially fixed by moving `parsingComputeLib` into `parsing/proofs`.
`Holmake -r cleanDeps` doesn't help and I am failing to reproduce this even on a fresh clone

formrre:
Fixes #575


xrchz:
This looks like a copy rather than a move. But either way, I am not yet on board with this change. I would like `parsingComputeLib` to _not_ depend on the parser `proofs`, because I expect people might want to execute the parser in the logic without building the proofs. (Also, I am not yet convinced that #575 describes a real problem on `master`.)

formrre:
I made a mistake and have not  deleted it. I have amended it. See the force push.

xrchz:
I'm going to close this pending further discussion on #575, because if there is a dependency problem, I don't think this is the right solution -- `parsingComputeLib` should not have to depend on proofs.

PR #577, opened 2018-12-07
Add 'local' blocks to semantics & compiler.

talsewell:
This adds the 'local .. in .. end' mechanism to the source AST, the parser & type system.

The backend is updated. This is pretty simple because 'local' disappears in the first source_to_flat pass.

There's more work to be done to support adding local declarations in ml_progLib, and to actually use the local mechanism to hide some parts of the basis that clutter things. However, given that there's a lot of things in flux, I'd like to merge the existing state as-is and work on progLib/translator implications for a subsequent merge.


xrchz:
Reopen when fixed

talsewell:
OK, translation is fixed, have re-merged with master, hopefully will succeed.

Issue #62, opened 2015-08-18
document, and possibly rethink, HOLHEAP policy


xrchz:
CakeML currently creates many Poly/ML heaps (one in most directories), building on previous heaps. The exact contents of each heap is a bit random, since there is no well-understood policy on what should be listed. Similarly for the parent heap, when present.

As I understand it, the point of heaps is to trade space for time. The point of this issue is to reevaluate our use of heaps, to pick a policy that is likely to save us the most time for common workflows on the repository, and to document that policy (probably at https://wiki.cakeml.org/Conventions).

myreen:
When working with the compiler proofs in compiler1, compiler1/semantics and compiler1/proofs, I find it very useful to have heaps, but to exclude the files that I'm currently working on from the heaps. By excluding the currently changing files, Holmake doesn't rebuild the heaps and loading times are good all around.

xrchz:
Is it possible to somehow get `Holmake` to automatically detect which files are changing, to get this good behaviour without having to edit the `Holmakefile`? Is there some HOLHEAP policy that would give us that behaviour?

Maybe `Holmake` could be made to only update the heap when told (i.e., when we are done editing), but otherwise default back to reloading changing files (but reusing the heap for non-changing files).

mn200:
If the files are already present in the heap, then reloading new versions of those same files over the top of them is not sure to work well.

xrchz:
One suggestion (from @tanyongkiam): everything should build _without_ heaps, and heaps should therefore only be an optimisation.

xrchz:
Here's a plan for action:
1. Remove all heaps and get the build to go through.
2. Measure any performance change (e.g., total regression time) to see if it's worth proceeding further.
3. Revive discussion here to settle on a transparent heap policy (i.e., optimisation only, no functional difference, as suggested above).
4. Implement it

xrchz:
The policy for now is "no heaps" unless anyone objects.

PR #583, opened 2018-12-14
Heapfree

xrchz:
Remove all HOLHEAPs and clean up `Holmakefile`s. Work on #62 .


xrchz:
This is looking like it's going to pass regression. Reviews from @mn200 and @myreen please?

basis/Holmakefile
```diff
@@ -1,19 +1,17 @@
-INCLUDES = ../developers ../misc pure ../translator ../characteristic ../semantics ../translator/monadic
-OPTIONS = QUIT_ON_FAILURE
+INCLUDES = $(CAKEMLDIR)/developers $(CAKEMLDIR)/misc\
+					 $(CAKEMLDIR)/semantics $(CAKEMLDIR)/characteristic\
+					 $(CAKEMLDIR)/translator $(CAKEMLDIR)/translator/monadic\
+					 pure
 
-all: $(DEFAULT_TARGETS) README.md
+CLINE_OPTIONS = --qof
```
mn200: Don't need / shouldn't include `--qof` anymore:

```suggestion
 
```

basis/Holmakefile
```diff
@@ -1,19 +1,17 @@
...
-TARGETS0 = $(patsubst %Theory.sml,,$(THYFILES))
-TARGETS = $(patsubst %.sml,%.uo,$(TARGETS0))
-
-all: $(TARGETS) basis_ffi.o README.md
+all: $(DEFAULT_TARGETS) README.md basis_ffi.o
 .PHONY: all
 
 README_SOURCES = $(wildcard *Script.sml) $(wildcard *Lib.sml) $(wildcard *Syntax.sml) $(wildcard *.c) dependency-order
 DIRS = $(wildcard */)
 README.md: $(CAKEMLDIR)/developers/readme_gen readmePrefix $(patsubst %,%readmePrefix,$(DIRS)) $(README_SOURCES)
```
mn200: If you're going to `protect` the file in the command, you should `dprot` it in the dependencies

basis/pure/Holmakefile
```diff
@@ -1,13 +1,8 @@
-INCLUDES = ../../misc $(HOLDIR)/examples/formal-languages/regular
-OPTIONS = QUIT_ON_FAILURE
+INCLUDES = $(CAKEMLDIR)/misc $(HOLDIR)/examples/formal-languages/regular
 
-all: $(DEFAULT_TARGETS) README.md
-
-THYFILES = $(patsubst %Script.sml,%Theory.uo,$(wildcard *.sml))
-TARGETS0 = $(patsubst %Theory.sml,,$(THYFILES))
-TARGETS = $(patsubst %.sml,%.uo,$(TARGETS0))
+CLINE_OPTIONS = --qof
```
mn200: Don't need `--qof`

```suggestion

```

candle/set-theory/Holmakefile
```diff
@@ -1,28 +1,9 @@
-INCLUDES = ../../misc $(HOLDIR)/src/pred_set/src/more_theories
-OPTIONS = QUIT_ON_FAILURE
+INCLUDES = $(HOLDIR)/src/pred_set/src/more_theories $(CAKEMLDIR)/misc
```
mn200: Don't need to include `more_theories`; these theories are in `sigobj` and always accessible

compiler/backend/closLangScript.sml
```diff
@@ -5,6 +5,8 @@
 *)
 open preamble backend_commonTheory;
 
+local open astTheory in end
```
mn200: Surely it should be backed_common that becomes the grammar ancestor.  Why is `ast` chosen here?

compiler/bootstrap/compilation/x64/64/Holmakefile
```diff
@@ -1,36 +1,16 @@
 ARCH = x64
 WORD_SIZE = 64
-INCLUDES = ../../../translation ../../../../../semantics ../../../.. ../../../../backend/$(ARCH) ../../../../encoders/asm ../../../../encoders/$(ARCH) ../../../../../basis
-CLINE_OPTIONS = --qof
+INCLUDES = $(CAKEMLDIR)/semantics $(CAKEMLDIR)/basis ../../../..\
```
mn200: Define `../../../../` in a Holmakefile variable (perhaps wrt `CAKEMLDIR` rather than current directory), and then reuse?

compiler/encoders/asm/asmSemScript.sml
```diff
@@ -3,6 +3,7 @@
 *)
 open HolKernel Parse boolLib bossLib
 open asmTheory machine_ieeeTheory
+open miscTheory (* for bytes_in_memory *)
```
mn200: Surely should be an `open preamble` here.

basis/Holmakefile
mn200: Grr; looks like proposed change would leave you with a line containing just `C`. How annoying.

compiler/backend/closLangScript.sml
xrchz: It's not a grammar ancestor in the sense of `set_grammar_ancestry`... But as for why I'm creating a dependency here, maybe see what happens if you remove it and try to build the file? (I'm not sure what will happen but I would guess the build fails.)

compiler/encoders/asm/asmSemScript.sml
xrchz: I agree, but I think we'll have to take that up with @myreen.

basis/Holmakefile
xrchz: Thanks for the suggestion. Is there any chance of getting Holmakefile includes or similar so the boilerplate lines for README.md generation can be written (and updated) in just one place?

compiler/bootstrap/compilation/x64/64/Holmakefile
xrchz: I've been following the convention that directories under the same top-level directory don't get the CAKEMLDIR treatment. And in this case, `$(CAKEMLDIR)/compiler` isn't shorter than `../../../..` so I'm going to leave it as is. (A separate shorter variable could be nice, but not worth it in my opinion.)

compiler/encoders/asm/asmSemScript.sml
xrchz: Absent @myreen's input, this stays as-is for now.

basis/Holmakefile
mn200: See HOL-Theorem-Prover/HOL#511

Issue #531, opened 2018-09-28
Incredibly large toX script files in bootstrap/translation


mn200:
Could we not break some of these up so that failing translations need not throw away large quantities of successful translation that preceded the error in super-large files?  Debugging errors in the bootstrap is otherwise extremely dispiriting.

If we had a long chain of smaller files, more progress would get saved.  Of course, the need to linearise is not ideal, but linear chains of massive files is particularly grim.

myreen:
Sounds like a good idea. Which files are particularly bad?

oskarabrahamsson:
I agree that this is useful. I think the overhead from loading/storing theories a few more times is definitely worth it. It would be nice have one toXScript for each pass, e.g. source_to_flat, flat_to_pat, pat_to_clos, etc. 

My personal experience is that all the 'interesting' stuff happens somewhere deep inside to_dataProgScript (clos, bvl or bvi), and running that interactively to the end just to fix a tiny thing takes quite some time.

oskarabrahamsson:
@lorenzleutgeb translation is linear so it has to be stuck in somewhere. If I recall correctly, lexerProg previously appeared before to_dataProg, but was moved down so that you would not have to build it just to go in and apply fixes to to_dataProg (which is much more likely to need adjustments when things change in the compiler).

xrchz:
Is this done now following #584?

oskarabrahamsson:
This was done in #584.

PR #584, opened 2018-12-19
Split to_dataProg into several theories

oskarabrahamsson:
This PR makes progress towards #531. (Much progress has already been made by @talsewell in PR #582.)

* `to_dataProgScript.sml` is split into six theories (one for each IR transition from source to dataLang).
* Some boilerplate still remains at the top of each file: code that changes the way the translator searches for constant definitions. Perhaps this is best solved by making the translator configuration more fine grained? Similar but different functions are needed later in the `{lexer,parser,to_word}Prog` theories.


talsewell:
Adding the "enhancement" label to test fetching said information from github. Had to be on an open PR. 
Don't mind me.

PR #589, opened 2018-12-23
Make bootstrapped compiler produce non-zero exit code on error

myreen:


myreen:
I get the impression that `master` fails in the same way, when tested on a machine with `Lem` installed. In other words, this failure in the regression test is not really a failure of this PR.

xrchz:
Feel free to run this on another machine / remove Lem from oven2 (although we do want to test Lem generation regularly) / test it privately.

PR #595, opened 2018-12-30
Local blocks in ml_progLib

talsewell:
This collection of patches adds local-block support to ml_progLib.

For this to be useful, this requires supporting a stack of open blocks, rather than the 1-2 available previously (global + one module). Nested modules are now supported and have been very very lightly tested. Along the way, the ML_code theorem shape had to change, and rather than trying to make sense of all the RATOR_CONVs and RAND_CONVs, I refactored the way that the relevant theorems are defined and applied.


talsewell:
Oops, closing this PR for now, there's a known failure, I haven't tracked down all the uses of ML_code, especially in characteristic.

talsewell:
I've fixed a few issues, let's find out what the regression test thinks.

PR #597, opened 2019-01-08
Lem 194778e

formrre:
Regenerates the scripts generated from lem files using the latest lem master and removes unused lem stubs.


formrre:
This looks like it has been regenerated correctly, but Holmakefiles is broken if lem setup is done differently. Probably abort the regression test, I need to make them more robust.

PR #604, opened 2019-01-15
Add header file for compcert compatibility

Gaj7:
Currently, compiling `basis_ffi.c` with compcert (the newest release, version 3.4) results in the following error "use of undeclared identifier 'S_IRUSR'". This macro, along with the other mode arguments for open(), are defined in `sys/stat.h`. Including this header resolves the issue.

I'm not sure why this isn't a problem when compiling under gcc. I assume one of the other header files includes `sys/stat.h`, but I was also under the assumption that both compcert and gcc would use the same system-provided header files. Regardless, the [linux man page](http://man7.org/linux/man-pages/man2/open.2.html) for open() seems to suggest including `sys/stat.h` (as well as `sys/types.h`, although this was not necessary on my machine), so including these may be wise for portability reasons.

### DIFF
--- a/basis/basis_ffi.c
+++ b/basis/basis_ffi.c
@@ -7,6 +7,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <fcntl.h>
+#include <sys/stat.h>
 #include <sys/time.h>
 
 /* clFFI (command line) */


xrchz:
Happy to merge this upon confirmation (from anyone) that the new file also works with gcc and/or clang on GNU/Linux and macOS

Gaj7:
I can confirm it works on GNU/Linux (specifically Ubuntu 18.04.1). Using gcc and then clang, I generated assembly for the original `basis_ffi.c`, and my modified version. I compared them with `diff`, and the only difference was the name of the file.

Unfortunately, I don't have the means to test on macOS.

PR #608, opened 2019-01-20
Replace cheats with ...

myreen:


xrchz:
I'm not very keen on this. What's the point? To help grep for cheats? The problems I see with this approach are that (a) it goes stale (people can introduce cheats instead of more ... by commenting things out), (b) it makes it harder to comment/uncomment while editing if you want to preserve this convention, and (c) it encourages the idea that one merely need grep for cheats rather than inspecting the oracles of theorems. You could address (a) and (b) by making the linter more powerful, but (c) would still be a problem absent lots of clear documentation...

myreen:
This is meant to be a one-off clean up before the next release. 

I don't think there should be a policy against syntactic cheats. If there was such a policy, then the linter should check for it, but there isn't such a policy and the linter is not checking this.

If I remember correctly, @SOwens wanted to not have cheats in the code base. I don't have a strong opinion, but thought it would be nice to not have syntactic cheats ahead of a release.

@xrchz you wrote "I'm not very keen on this". Is this really saying "I'm against this" or is it more like "I think this is pointless and don't care"?

xrchz:
I meant against rather than don't care, but less so if it's just a one off. I won't stop this PR if folks want it.

PR #621, opened 2019-03-21
astToSexprLib: fix printing of nil strings

michaelsproul:
Previously the unverified sexpr printer would produce malformed s-expressions for terms involving anonymous patterns/variables, like this one:

```sml
(fn (x, y) => g x y)
```

This would produce `(Fun nil (Mat (Var (Short nil))) ..)` rather than `(Fun "" (Mat (Var (Short ""))) ..)`.


xrchz:
This whole file should be using Syntax libraries (astSyntax, and stringSyntax for the new change). But this change is fine given the state of the rest of the file.

PR #623, opened 2019-03-29
Linear scan: faster version

TWal:
This pull request features a faster version of the linear scan register allocator.
It uses monadic arrays instead of num_map and lists in several places.

It is now about 2 times faster than the coalescing register allocator without coalescing (reg-alloc=0)


myreen:
These update have been merged into the `monadic-trans-cleanup` branch, where the issue with the bootstrap translation has been resolved. I'm closing this PR.

PR #625, opened 2019-04-08
Fix for removal of script-file new_attribute fn

mn200:


myreen:
This was made redundant by #628

PR #629, opened 2019-04-10
Merge vstte18 branch

hferee:
Most changes include a more accurate file system model.
iocat has also been added back, and is now implemented using a TextIO.copy function, similar to ocaml's BatIO.copy.


xrchz:
This looks good at a skim - just need to fix the regression errors.

hferee:
It looks fine on my side now, you can start the regression test again.

PR #630, opened 2019-04-12
Update backendComputeLib use of ThmSetData.

talsewell:
The obvious updates to use the new interface.

I've tried to test it. I can load it and run its function. I've tried to get to the examples that use it,
but something is up in basis/ListProgScript which I don't think is related. So this should be progress
towards fixing master but I'm not even sure of that.

### DIFF
--- a/compiler/backend/backendComputeLib.sml
+++ b/compiler/backend/backendComputeLib.sml
@@ -17,6 +17,9 @@ end
 
 open Parse
 
+fun theory_computes thy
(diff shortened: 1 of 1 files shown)


mn200:
The failure in ListProg is fixed by 1754dcb2ad which I pushed directly to `master`.  I hope the regression framework will (can be made to) retry this PR with that more recent version of master.  (The regression test first used, № 788: https://cakeml.org/regression.cgi/job/788 has gone into a loop on `ListProg` already.)

Issue #631, opened 2019-04-13
Fix ARM8 bootstrap cross compile


tanyongkiam:
The (unverified) bootstrap sexpression can be cross-compiled on x64 for all of the backends successfully except ARM8.

This is probably because the loc_offset range on ARM8 needs to be extended.

This issue is about fixing that (which I will try) and seeing if that fixes the cross compilation.

tanyongkiam:
I'm considering this closed because the cross compile now succeeds.

Note it requires passing the flag --jump=false.

PR #632, opened 2019-04-22
Extends loc_offset on ARM8 to 32bits

tanyongkiam:
An attempt at fixing #631


tanyongkiam:
Closing for now because this doesn't fully resolve #631

tanyongkiam:
After some investigation, this actually does mostly resolve #631 

In particular, the ARM8 cross compile bootstrap now succeeds after passing --jump=false to the default ARM8 configuration (which has that flag set to true).

The regression test for this PR already completed successfully.

PR #646, opened 2019-05-12
add map2 mapping

larsrh:
Surprisingly, this was the only change required on current master to export Lem to Isabelle.


xrchz:
Presumably this should be accompanied by a corresponding change to `libScript.sml`?

larsrh:
Why? This doesn't result in any change there, or does it?

xrchz:
Perhaps not. I'm happy to merge if it doesn't - would you mind confirming?

PR #657, opened 2019-05-27
Add support for (double) floating-point arithmetic to the translator

### LINKED ISSUE #636: Extend translator to support floating-point operations

HeikoBecker:
This adds the appropriate theorems and automation to `ml_translatorScript`/ `ml_translatorLib.sml` to support floating-point arithmetic in the translator.
There are also 2 simple tests in `ml_translator_testScript.sml`.
Closes #636


translator/ml_translatorLib.sml
```diff
@@ -424,9 +424,9 @@ val basic_theories =
    ["alist", "arithmetic", "bag", "bitstring", "bit", "bool",
     "combin", "container", "divides", "fcp", "finite_map", "float",
     "fmaptree", "frac", "gcdset", "gcd", "ind_type", "integer_word",
-    "integer", "integral", "list", "llist", "marker", "measure",
-    "numeral_bit", "numeral", "numpair", "numposrep", "num", "one",
-    "operator", "option", "pair", "path", "patricia_casts",
+    "integer", "integral", "list", "llist", "marker", "machine_iee",
```
xrchz: This looks wrong... doesn't "machine_iee" need another "e"? Why wouldn't the regression test catch this?

HeikoBecker: Interesting. Here is my "suggested fix":
```suggestion
    "integer", "integral", "list", "llist", "marker", "machine_ieee",
```

Could it be that it is not caught by the regression test because the variable is only used in function `full_name_of_type` below, which searches for the `thy_name` if it is not in the list?

xrchz:
Thanks for fixing the typo. Why were changes to this list of "basic theories" required anyway? I think it's a bad sign that there's a hardcoded list of theory names in the translator. Maybe @myreen could comment on whether the changes to this list make sense or not?

HeikoBecker:
The change was not required. I saw the list of basic theories and my assumption was that it needs to be extended.

PR #659, opened 2019-06-06
Update filterProg for methods renamed upstream

IlmariReissumies:


IlmariReissumies:
That's a quick merge! To be clear, I never actually tested this change, but I suppose it's innocent enough.

PR #660, opened 2019-06-11
Change assignments to equalities in assert statements

Gaj7:
I believe `assert(clen == 8)` was meant instead of `assert(clen = 8)`. The former asserts that `clen` is 8. The latter _mutates_ `clen`, and the assertion always holds.

Also, I caught this by compiling with the `-Wall` gcc flag. Perhaps it would be useful to add this to the Holmakefile which builds this file?

### DIFF
--- a/basis/basis_ffi.c
+++ b/basis/basis_ffi.c
@@ -99,7 +99,7 @@ void ffiopen_out (unsigned char *c, long clen, unsigned char *a, long alen) {
 }
 
 void ffiread (unsigned char *c, long clen, unsigned char *a, long alen) {
-  assert(clen = 8);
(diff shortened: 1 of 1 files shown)


xrchz:
Thanks for this! If `-Wall` is reasonably portable, consider submitting that change in this or another PR?

Gaj7:
By portable, do you mean accepted by different c compilers? I know gcc, clang, and compcert all accept the flag and interpret it as "turn on common warnings", but the specific warnings may be compiler specific, I'm not sure.

xrchz:
Yes accepted by most compilers is all I meant.

PR #678, opened 2019-08-12
Fix semantics/proofs for new drule behaviour

talsewell:
This branch will contain fixes for the new drule behaviour.

I've started a PR so that it regression tests and can be merged to master once it builds. It's likely that it won't be done for a while though.


talsewell:
Closed for now, not finished, Michael is working on another approach, and it's hurting the regression infrastructure.

PR #688, opened 2019-09-07
Close issue 683

myreen:


myreen:
The regression test hit the same [failure that's on master](https://cakeml.org/regression.cgi/job/1028). I have separately tested that the bootstrap translation goes through.

Issue #675, opened 2019-08-01
Rename CompileError to AssembleError


myreen:
During a discussion on slack today, @SOwens and @xrchz suggested that `AssembleError` would be a more fitting name for `CompileError`. I agree. 

This issue is about renaming `CompileError` to `AssembleError`.

LawAbidingCactus:
I've made this change in #693.

xrchz:
Closed by #693

PR #693, opened 2019-10-06
rename CompileError to AssembleError

LawAbidingCactus:
Renames CompileError to AssembleError, as discussed in #675.

### DIFF
--- a/compiler/compilerScript.sml
+++ b/compiler/compilerScript.sml
@@ -64,7 +64,7 @@ val _ = Datatype`
      ; only_print_sexp     : bool
      |>`;
 
-val _ = Datatype`compile_error = ParseError | TypeError mlstring | CompileError | ConfigError mlstring`;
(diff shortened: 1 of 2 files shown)


xrchz:
Awesome, thanks! Will merge if it gets through the regression test.

compiler/compilerScript.sml
```diff
@@ -128,7 +128,7 @@ val error_to_str_def = Define`
        concat [strlit "### ERROR: type error\n"; s; strlit "\n"]
      else s) /\
   (error_to_str (ConfigError s) = concat [strlit "### ERROR: config error\n"; s; strlit "\n"]) /\
-  (error_to_str CompileError = strlit "### ERROR: compile error\n")`;
+  (error_to_str AssembleError = strlit "### ERROR: compile error\n")`;
```
xrchz: Can we change the error message too to say assembly or assemble instead of compile?

LawAbidingCactus: Done!

PR #695, opened 2019-10-30
Ocaml infixes

mn200:


mn200:
This isn't worth testing until HOL's commit HOL-Theorem-Prover/HOL@87e453f makes its way to HOL `master`.  (I do hope that this will be soon ; maybe in < 24 hours.)

Issue #699, opened 2019-11-09
cfLetAutoLib is sensitive to load order


oskarabrahamsson:
`cfLetAutoLib` behaving differently depending on when it is loaded. This issue is about making the code robust against that.

Here is an example: the simpset `DEF_SIMPSET` is used by some of the library internals. This simpset is `srw_ss` taken at some point in time, plus whatever you add to it later using the interface. Depending on when this code is run, `xlet_auto` will know different things. [Here is the relevant code](https://github.com/CakeML/cakeml/blob/dc73bd8f21601e06bdc92eb303865a3992673970/characteristic/cfLetAutoLib.sml#L220):
```sml
val DEF_SIMPSET = ref pure_ss;
val _ = (DEF_SIMPSET := srw_ss());

(* TODO: Find a way to export that - like with ThmSetData.new_exporter *)
fun add_simp_frag sf = (DEF_SIMPSET := ((!DEF_SIMPSET) ++ sf));
fun get_default_simpset () = !DEF_SIMPSET;
```

There may be other things like this.

mn200:
I'd say that DEF_SIMPSET should be built on a fixed base (`list_ss` perhaps) and then augmented by a fixed set of fragments grabbed by using `BasicProvers.thy_ssfrag`.

konrad-slind:
I agree. Things that depend on stateful simpsets are fragile.

oskarabrahamsson:
@mn200 Thanks for the suggestion.

This issue can be closed once #703 is merged.

PR #700, opened 2019-11-09
Patch up broken proof in iocatProg (see comments)

oskarabrahamsson:
This is just a quick fix to get master building again.

The real issue here is that various rewrites and other things done
internally in cfLetAutoLib seems to depend on _when_ cfLetAutoLib
was loaded. When the Lib is loaded appears to have changed after a
recent update Holmake. This makes the xlet_auto heuristics instantiate
some variables in some theorems with the wrong things.

### DIFF
--- a/examples/iocatProgScript.sml
+++ b/examples/iocatProgScript.sml
@@ -30,18 +30,18 @@ val catfiles_string_def = Define`
     concat (MAP (λfnm. file_contents fnm fs) fns)
 `;
 
-val cat_spec0 = Q.prove(
-  `∀fns fnsv fs.
-     LIST_TYPE FILENAME fns fnsv ∧
-     EVERY (inFS_fname fs) fns ∧
-     hasFreeFD fs
-    ⇒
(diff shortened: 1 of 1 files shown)


mn200:
Libraries should not depend on load order at all, so the ideal fix would be to `cfLetAutoLib`...

xrchz:
I agree with Michael. Why does the library depend on when it's loaded?

xrchz:
I see there is issue #699 to address that, so this pull request can go ahead when it passed. Thanks Oskar!

oskarabrahamsson:
I also agree. Until whatever is up with `cfLetAutoLib` is fixed it would be nice if a few PRs could build, though.

I wrote an issue about this: #699. `cfLetAutoLib` picks out `srw_ss` at some point in time. The tactic succeeds or fails with various goals depending on what's in the simpset at the time when the library is loaded. I have no idea how it works so I can't say what simplifications it should know about.

myreen:
The regression test failed, but this is an improvement on master so I will merge it.

PR #704, opened 2019-11-12
Add CAKE_BIN as a dependency to cake.S in unverified bootstrap

oskarabrahamsson:
The binary used in the unverified bootstrap was not being downloaded when building `cake.S`, because it was not listed as a dependency. This fixes this issue.

~~Building the rest of the unverified bootstrap will still get stuck though, because of HOL-Theorem-Prover/HOL#752.~~


oskarabrahamsson:
Opening this when it's all done.

xrchz:
I would like the addition of `exec` removed, given that the bug that requires it has been fixed in HOL.

oskarabrahamsson:
I don't see why but there you go.

Issue #705, opened 2019-11-18
The translator accepts constants with preconditions


oskarabrahamsson:
The translator accepts constants with preconditions, and duplicates preconditions:

```sml
> val foo_def = Define` foo = 5 DIV 0`;

Definition has been stored under "foo_def"
val foo_def =  [] ⊢ foo = 5 DIV 0: thm
```

```sml
> val th = translate foo_def;

Translating foo

WARNING: foo has a precondition.

val th =  [foo_side, PRECONDITION foo_side] ⊢ NUM foo foo_v: thm
```

What's worse, the precondition also leaks into the theorem produced by `ml_progLib`:
```sml
> get_ml_prog_state() |> ml_progLib.get_thm;
val it =  [foo_side] ⊢ ... ... [... ] (scratch_st ffi): thm
```

PR #706, opened 2019-11-18
Translate validAddSym before cmlPEG

oskarabrahamsson:
Allowing `cmlPEG` to have a precondition propagates the precondition all the way to the top-level compiler theorem because of an issue with the translator (#705).


mn200:
Huh.  In an ideal world, should my fix in 2680f32a6 have worked?

oskarabrahamsson:
Ideally the translator would 
* not put `cmlpeg` into the ml program before the side condition had been removed; or
* refuse to translate `cmlpeg` until you gave it a theorem about `validAddSym` so that it could get rid of the precondition.

Also, `ml_progLib` should probably refuse all v_thm that has non-empty hypotheses.

PR #707, opened 2019-11-26
Define space cost semantics and add new top-level compiler correctness theorem

myreen:
The cost semantics is based on dataLang. If one proves that the program is safe for space, then one gets a stronger compiler correctness result, namely:

https://github.com/CakeML/cakeml/blob/1e04d1b7c37b6fa205a8106c64c00ddf5050b0aa/compiler/proofs/compilerProofScript.sml#L273-L282


tanyongkiam:
The specific theorem that one might use in practice is:
https://github.com/CakeML/cakeml/blob/1e04d1b7c37b6fa205a8106c64c00ddf5050b0aa/compiler/backend/proofs/backendProofScript.sml#L3483

PR #714, opened 2020-01-06
Replace packLib's custom theorems with ThyDataSexp

mn200:
This has no huge advantage except that it doesn't contaminate the theorem database with encoded state (which then show up as results to `DB.match` *etc*).  It's nice also to use standard HOL APIs.

Some uses of `packLib` were entirely removed in favour of just quoting terms and types in `Lib.sml` files directly, which can be done safely with local rebinding of the `Parse` structure.


myreen:
Bootstrap translation took more than 17 hours rather than the expected 3-4 hours. I therefore mark the current state as _test failing_.

mn200:
Can you set up a fresh regression test for this branch please?  My local tests indicate it to now be faster than `master`.

PR #715, opened 2020-01-06
Add unsafe array access to CakeML source language

myreen:
An unsafe array access is an array access that crashes if the index is out of bounds. Note that these new unsafe primitives do not have a type in the type system. As a result, the type checker disallows programs containing such accesses.

These new accesses can be used in programs that are verified, if the type checker is by-passed. They can e.g. be used in CF verified programs and (in the future) programs produced by the monadic translator.


myreen:
I'm closing this for now because I realise that more work is needed.

PR #718, opened 2020-01-13
LPR proof checker

tanyongkiam:
This updates the LRAT proof checker to use LPR format


myreen:
I marked this as test failing because this PR contains the commits of the `unsafe-accesses` branch. Those currently break the bootstrap translation. (I'm fixing that today.)

PR #721, opened 2020-01-28
tweak basis_ffi.c around exit and cml_exit

tanyongkiam:
- Runtime.exit now exits directly instead of redirecting to cml_exit

- adds a (on by default) flag that prints a message to stderr when memory exhaustion exits happen in cml_exit


tanyongkiam:
Ah, that's a good point. I think those things should probably have used Runtime.abort or Runtime.exit instead.

PR #728, opened 2020-02-21
Comment cheated theorem `ascending_infinite_suffix`

arolle:


arolle:
Master will not fail, as no `check_thm` is around `ascending_infinite_suffix`, thus closing.

PR #735, opened 2020-05-08
LPR example updates

tanyongkiam:
This adds some error messages and other updates to the LPR proof checker (+ its proofs).

Most notably, it adds another "entry point" whole_prog_spec2 (analogous to whole_prog_spec).

I believe whole_prog_spec2 should replace all existing uses of whole_prog_spec (and then the names can be updated).

=== For regression testing, do not merge ===


tanyongkiam:
It seems like this worked https://cakeml.org/regression.cgi/job/1222

And the latest failure is unrelated.

PR #738, opened 2020-05-25
Export update

tanyongkiam:
Do not merge, for regression testing purposes only.

This updates the export scripts to use a CakeML "main" function.


myreen:
I think it's likely that this gets past the regression test soon, e.g. [the currently running job](https://cakeml.org/regression.cgi/job/1291).

I see that reviews from @xrchz and @IlmariReissumies are still pending. Are there any objections to merging this in case it passes the regression test?

basis/basis_ffi.c
```diff
@@ -20,8 +20,14 @@
 /* clFFI (command line) */
 
 /* argc and argv are exported in cake.S */
```
xrchz, line 22: is this still true?

basis/basis_ffi.c
```diff
@@ -247,3 +253,62 @@ void ffidouble_toString (unsigned char *c, long clen, unsigned char *a, long ale
   // for the 0 byte
   assert (bytes_written <= 255);
 }
+
+void main (int largc, char **largv) {
```
xrchz, line 257: it may be clearer to write "local_" instead of just "l"

basis/basis_ffi.c
```diff
@@ -247,3 +253,62 @@ void ffidouble_toString (unsigned char *c, long clen, unsigned char *a, long ale
...
+
+  argc = largc;
+  argv = largv;
+
+  char *heap_env = getenv("CML_HEAP_SIZE");
+  char *stack_env = getenv("CML_STACK_SIZE");
+  char *rest;
```
xrchz, line 264: Why is this called "rest"? (Maybe "s" or "temp"?)

unverified/sexpr-bootstrap/x64/32/Holmakefile
```diff
@@ -16,7 +14,7 @@ README.md: $(CAKEMLDIR)/developers/readme_gen readmePrefix $(patsubst %,%readmeP
 cake-sexpr-$(ARCH)-$(WORD_SIZE): *$(ARCH)SexprScript.sml
 
 cake.S: cake-sexpr-$(ARCH)-$(WORD_SIZE) $(CAKE_BIN)
-	$(CAKE_BIN) --sexp=true --exclude_prelude=true --skip_type_inference=true --emit_empty_ffi=true --heap_size=$(HEAP_MB) --stack_size=$(STACK_MB) --reg_alg=0 <$< >$@
+	CML_STACK_SIZE=8000 CML_HEAP_SIZE=8000 $(CAKE_BIN) --sexp=true --exclude_prelude=true --skip_type_inference=true --emit_empty_ffi=true --reg_alg=0 <$< >$@
```
xrchz, line 17: could these numbers (8000) not still be Holmakefile variables (as previously)? also why did they change?

tanyongkiam: This 8000 here is specifically to set the CakeML compiler's heap and stack (of $(CAKE_BIN)) when bootstrapping itself using the sexpression representation.

mn200: I guess @xrchz means you could still write
```
CML_STACK_SIZE = 8000

cake.S: ....
    CML_STACK_SIZE=$(CML_STACK_SIZE) ....
```

Perhaps it'd be nicer still if Holmake supported the `export` directive that normal `make` has; then you'd write

```
export CML_STACK_SIZE = 8000
```

and wouldn't need to use the prefix in the recipe at all.

IlmariReissumies:
This is quite an improvement! I like how the assembly template gets simpler.

My use case for this would be to completely replace the `main` function with something else, which seems easy enough to do now: just make sure the `cml_*` variables are initialised correctly in `main`. To do this in a principled way, I wonder if the assumptions are documented anywhere? For example, are there alignment requirements for the three `cml_*` pointers? Is it required that the stack starts where the heap ends? A comment or two about this in (say) `basis_ffi.c` would be nice.

In any case, I'd be happy to see this merged with or without that.

PR #745, opened 2020-08-13
Associativity Rewriting Correctness Proof

rodalfus:
This pull request implements the following:
- Add correctness proof for the associativity rewriting
- Remove previous unused and commented partial proof


icing/icing_optimisationProofsScript.sml
```diff
@@ -603,11 +603,376 @@ Proof
   \\ fs[]
 QED
 
+fun rename_each [] = ALL_TAC
+| rename_each [first] = qmatch_asmsub_rename_tac first \\ qpat_x_assum first mp_tac \\ disch_then assume_tac
```
HeikoBecker, line 607: Is there a particular reason why you are using `qmatch_asmsub_rename_tac` here instead of `rename1`?

icing/icing_optimisationProofsScript.sml
```diff
@@ -603,11 +603,376 @@ Proof
...
+| rename_each (first::rest) = qmatch_asmsub_rename_tac first \\ qpat_x_assum first mp_tac \\ rename_each rest \\ disch_then assume_tac
```
HeikoBecker, line 608: `disch_then assume_tac` -> `strip_tac` for brevity

icing/icing_optimisationProofsScript.sml
```diff
@@ -603,11 +603,376 @@ Proof
...
+
 Theorem fp_assoc_gen_correct:
   ∀ fpBop st1 st2 env e r.
     is_rewriteFPexp_correct [fp_assoc_gen fpBop] st1 st2 env e r
 Proof
-  cheat
+  rw[is_rewriteFPexp_correct_def]
+  \\ REVERSE (qspecl_then [‘e’, ‘fpBop’] strip_assume_tac fp_assoc_gen_cases)
+  >- (
+  fs[]
```
HeikoBecker, line 618: Can you remove the line break after the `fs[]`? It is usually a good practice to put "conceptual steps" on a single line, but there is no need for `fs[]` to be "alone"

HeikoBecker: Similarly for snippets like
    
    rveq
    \\ fs[]

and other below. They should be streamlined into a single line.

icing/icing_optimisationProofsScript.sml
```diff
@@ -603,11 +603,376 @@ Proof
...
+  \\ qpat_x_assum ‘evaluate _ _ [e2] = (_, _)’ mp_tac
+  \\ rename1 ‘evaluate st1 env [e3] = (st1N, Rval [v3])’
+  \\ disch_then assume_tac
+
+  \\ simp[do_app_def, CaseEq"option", CaseEq"v", CaseEq"prod", CaseEq"result"]
+  \\ rpt strip_tac
+  \\ rveq
+  \\ fs[]
+  \\ rveq
+  \\ imp_res_tac evaluate_sing
```
HeikoBecker, line 661: This line is a good example for what I meant above with "conceptual steps". You have an assumption `evaluate _ _ [e] = (_, v)` and want to prove that `v` must be a singleton list. Thus `imp_res_tac evaluate_sing \\ rveq \\ fs[]` can be put onto a single line (in my opinion)

icing/icing_optimisationProofsScript.sml
```diff
@@ -603,11 +603,376 @@ Proof
...
+  \\ fs[]
+  \\ rename1 ‘evaluate _ env [e1] = (st3N, Rval [v1])’
+  \\ rveq
+
+  (* Do the renaming dance again *)
+  \\ rename1 ‘evaluate st1N env [e2] = (st2N, Rval [v2])’
+  \\ qpat_x_assum ‘evaluate _ _ [e2] = (_, _)’ mp_tac
+  \\ rename1 ‘evaluate st1 env [e3] = (st1N, Rval [v3])’
+  \\ disch_then assume_tac
+
```
HeikoBecker, line 672: Can you remove the intermediate linebreaks? If they are supposed to split up the proof, please add comments, otherwise they "clutter up" the script.

HeikoBecker:
Overall, nice work. Thanks for doing the proof! Once the minor nitpicks are resolved, this is ready for merging.

icing/icing_optimisationProofsScript.sml
rodalfus: Yes; `rename1` also looks for matches in the goal. In this case, we only want to rename assumptions.

icing/icing_optimisationProofsScript.sml
mn200: In this case, they are probably equivalent (because what is being stripped was originally an assumption), but note that `disch_then assume_tac` and `strip_tac` are not always equivalent.  In particular, they behave differently on goals such as `(?x. P x) ==> Q`, `p /\ q ==> r` and `p \/ q ==> r`.

HeikoBecker: Thanks for pointing this out!

HeikoBecker:
Thanks for the addressing all the nitpicks.

PR #746, opened 2020-08-26
Model-theoretic conservativity for overloading version of Candle

IlmariReissumies:
This PR replaces the previous monolithic model construction with a more incremental construction by @arolle that retains parts of the old model when a theory update happens. If the theory update introduces overloads, we must discard the old model for those constants and types that depend on the new overloads. This allows us to prove a variation of model-theoretic conservativity, and obtain consistency as a corollary.


myreen:
I've marked this PR as "test failing" because the regression job seems to go into an infinite loop in `candle/overloading/syntax`. I wonder if a proof script there might not have been tested with the latest HOL master?

IlmariReissumies:
Thanks @myreen! Sorry, I had only tested `candle/overloading/semantics` locally, forgetting that the `syntax` has things that aren't used in `semantics`... everything builds for me locally now, on the same HOL commit as the failing regression test.

@arolle there was a failing proof in `holSyntaxCyclicityScript` (`ascending_infinite_suffix`) that I commented out for now because it looks like WIP, and parts of it were cheated already. I think this should be fine for now because it's not yet used for anything, and no `check_thm` depends on it. Do you agree or should we do something more sophisticated for this theorem?

arolle:
Thanks @IlmariReissumies  for the fixes. Indeed, some of the `syntax` development should ideally not have been part of this PR and is WIP, so I suggest to keep it `cheat`ed for now.

myreen:
The regression test got past all the relevant parts, so I'm merging this. 

By the way, `model-conservative-ext-ConstSpec-wit` is quite the branch name! :-)

PR #751, opened 2020-09-12
PIC-compatible x64 cml_main

### LINKED ISSUE #438: SEGFAULT on Alpine Linux
When I built it with the makefile I got a segfault

```
$ ./cake
Segmentation fault
```

but if I built it with `-no-pie -fno-PIE` it works.

I am on alpine linux with musl libc. I think the problem is related to TEXTRELs but I am not an expert.

Thanks to nsz for suggesting the no pie flags.

sorear:
movabs cannot be used in position-independent code without text
relocations as it embeds an absolute address into the text segment.
This was originally added to support heaps+stacks greater than 2GiB, but
since 01f9fa8e the heap and stack are no longer in the data segment and
the data segment size is negligible.

Fixes #438.

(I have tested the assembly code in this patch in an Alpine container with a previously built cake.S.  I have not yet run the bootstrap with the HOL patch itself.)

### DIFF
--- a/compiler/backend/x64/export_x64Script.sml
+++ b/compiler/backend/x64/export_x64Script.sml
@@ -28,17 +28,14 @@ val startup =
        "     .globl  cdecl(cml_stack)";
        "     .globl  cdecl(cml_stackend)";
        "cdecl(cml_main):";
-       "     pushq   %rbp                      # push base pointer";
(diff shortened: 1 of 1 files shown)


tanyongkiam:
This change looks good to me (essentially, reverting 7982b7).

I tested this on a local assembly file I had and it works fine (Ubuntu 18.04)

PR #752, opened 2020-09-12
two riscv fixes

sorear:
1. `cml_heap` was not being correctly dereferenced in the export code, so programs were starting with a 1 word heap and stack and immediately crashing with an out of memory error.
2. `tp` was being allocated as a general register, causing crashes at program termination when glibc tried to access thread local storage.  Added it to the avoid list.

Tested by building a `--target=riscv` compiler from `unverified/sexpr-bootstrap`, then running that compiler under qemu to generate a second compiler identical to the first; also ran the proof scripts under `compiler/proofs` and `compiler/backend/riscv/proofs`.

### DIFF
--- a/compiler/backend/riscv/export_riscvScript.sml
+++ b/compiler/backend/riscv/export_riscvScript.sml
@@ -41,8 +41,7 @@ val ffi_asm_def = Define `
   (ffi_asm (ffi::ffis) =
       SmartAppend (List [
        strlit"cake_ffi"; implode ffi; strlit":\n";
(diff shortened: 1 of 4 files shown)


tanyongkiam:
Thanks for cleaning these up! I might have made the same errors for the other targets wrt dereferencing `cml_heap`.

There's a comment here https://github.com/CakeML/cakeml/blob/e1e45b8b33b7e880159da4f7e1cac69d6252863d/compiler/backend/riscv/riscv_configScript.sml#L15 that says "not sure if they need to be avoided", would you mind deleting that comment?

PR #791, opened 2020-09-28
Optimize LPR checker

tanyongkiam:


myreen:
The regression test got past all the affected files, so I'll merge this.

Issue #797, opened 2020-10-15
improving access to global variables


sorear:
Right now a read or write of a global slot produces something like:

```
  a6bdc4:       be e0 10 00 00          mov    $0x10e0,%esi
  a6bdc9:       49 8b 7d c0             mov    -0x40(%r13),%rdi
  a6bdcd:       48 d1 e6                shl    %rsi
  a6bdd0:       4c 89 f0                mov    %r14,%rax
  a6bdd3:       48 01 c6                add    %rax,%rsi
  a6bdd6:       48 c1 ef 09             shr    $0x9,%rdi
  a6bdda:       48 89 f0                mov    %rsi,%rax
  a6bddd:       48 01 f8                add    %rdi,%rax
  a6bde0:       48 8b 48 08             mov    0x8(%rax),%rcx
```

```
  430af8:       fffff337                lui     t1,0xfffff
  430afc:       ed434313                xori    t1,t1,-300
  430b00:       fc0d3383                ld      t2,-64(s10)
  430b04:       00131313                slli    t1,t1,0x1
  430b08:       01bde433                or      s0,s11,s11
  430b0c:       00830333                add     t1,t1,s0
  430b10:       0093d393                srli    t2,t2,0x9
  430b14:       00730333                add     t1,t1,t2
  430b18:       00833383                ld      t2,8(t1) # fffffffffffff008 <__global_pointer$+0xffffffffff1b1fe0>
```

The first obvious issue is that the shift of a constant could be simplified.  Once that happens, the offset can be combined with the load, saving 2 instructions (riscv/arm large offset) or 3 (x64 case where it can be folded completely); there are 3197 accesses to -0x40(%r13) in the current compiler.

A naive search for shifted constants with no intervening instructions produces 12727 hits.  A more detailed analysis of the benefit here remains to be done.

There are 2155 calls to AllocGlobal at 32 bytes each, which could be in a loop.

We could also cache the real_address of the globals array in an additional store slot, since it only moves when the GC moves it and the GC can keep it updated.  This would reduce global accesses _to_ 2 instructions.  For riscv64 and to a lesser extent arm7 it might make sense to have several copies of the slot with separate shifted copies of the pointer because each can only address 4kB.  A complementary approach most useful on arm8 would be to keep the global pointer in a real register; up to 4096 globals could be accessed with a single register.

myreen:
[...]

PR #802, opened 2020-10-15
Latest data cost work

myreen:


sorear:
Is ElemAt just an optimization or is it semantically needed?  For #797 I was thinking of adding a special case to real_offset, which seems marginally better (handles Update as well) and much less invasive (no semantic changes, keeps constructor count in bvl/bvi/data down).

myreen:
At present it is just an optimization; it is the same as `El` but with one argument inlined to be an immediate. This reduces code bloat at the DataLang level. Previously we were relying on constant folding at the WordLang level to inline the immediate value. 

I have a vague plan to make the semantics of `El` and `ElemAt` differ. Currently they can both access Block values and ValueArrays. I'm considering to make `ElemAt` only able to access Block values, which are immutable unlike ValueArrays. By allowing `ElemAt` to only access Block values, an in-development common subexpression elemination pass at bvi_to_data could be more aggresive with `ElemAt`.

sorear:
Ah, so this is completely independent of #797 because that deals with Globals, which is always a ValueArray.

PR #805, opened 2020-11-16
Add Char.some and use it in TextIO

myreen:


myreen:
There is more work to come on this one, so I'll close this PR for now.

sorear:
Rather than changing the source semantics and everything that touches them, why not peephole optimize char <-> int <-> word8 conversions in one of the passes have to modify anyway?

tanyongkiam:
The specific pattern that inspired this was `Char.chr (Word8.toInt ...)` at the source level. It might be a little specific to catch this pattern if the functions are not yet inlined to the underlying CakeML primitives.

sorear:
This PR seems to be exploiting the ability of the translator to directly generate primitives though - it could generate WordFromInt/WordToInt and Ord/Chr primitives?

tanyongkiam:
I think you're right, but I think the translated basis generates something roughly like this at the source (I might be mistaken):

`val Char.chr = Chr;  ... ; val _ = Char.chr (...)`

So I think there is an indirection to lookup `Char.chr` first in the way of the peephole optimization.

translator/ml_translatorLib.sml
```diff
@@ -2865,7 +2865,9 @@ val builtin_monops =
    Eval_empty_ffi,
    Eval_force_out_of_memory_error,
    Eval_Chr,
-   Eval_Ord]
+   Eval_Chrw,
+   Eval_Ord,
+   Eval_Ordw]
```
sorear, line 2870: I don't see why this, specifically, can't generate a Chr (WordToInt) at the level of primitives.  I am not talking about the basis.

tanyongkiam: Ah I see what you mean now, so having `Char.chrw` translate to `Chr(WordToInt)` -- that makes sense, I misunderstood earlier.

myreen:
@sorear is completely right. I shouldn't add this to the source, instead the compiler should be smart enough to spot the patterns `word8->int->char` and `char->int->word8` to avoid the range checks that happen when converting `int->char` and `int->word8`.

I'll close this PR for now.

PR #808, opened 2020-12-01
Add Char.some and use it in TextIO -- Attempt 2

myreen:
This is a continuation of [Add Char.some and use it in TextIO](https://github.com/CakeML/cakeml/pull/805) #805, but forks of before the unproductive commits.


myreen:
I've done a manual build and this is fine. The regression test failure was just a glitch on oven2.

PR #816, opened 2021-02-16
Theory for Cyclicity check

arolle:
This proves an equivalence theorem for dependencies of overloading definitions, that states that cyclic overloading definitions have a certain shape.
This theorem is Lemma 5.17 of Kunčar's paper ["Correctness of Isabelle's Cyclicity Checker: Implementability of Overloading in Proof Assistants"](https://dl.acm.org/doi/10.1145/2676724.2693175).

The proofs in this PR serve as the theoretical foundation for a cyclicity checker for Candle with overloading definitions.


myreen:
Awesome work!

PR #823, opened 2021-03-28
Adapt arm8 backend for Apple platforms

oskarabrahamsson:
This PR consists of changes to the arm8 backend configuration and export scripts, enabling CakeML to produce binaries that can run on Apple arm64 platforms. This is achieved by the following two minor changes:
- avoid using the r18 register in the arm8 backend
- use relative addressing in the arm8 export glue code

See [here](https://developer.apple.com/documentation/xcode/writing_arm64_code_for_apple_platforms) for some Apple-specific divergences from the ARM64 specs. According to [some official ARM docs](https://developer.arm.com/documentation/ihi0055/d/), r18 is platform-reserved and should not be used in platform-independent code if possible. 

I have tested by compiling and running binaries for the OpenTheory checker on both macOS and Ubuntu ARM64 (under QEMU on the same machine), and everything appears to be working, but I have no idea what I'm doing, so some review of this PR would be appreciated.


tanyongkiam:
The arm8 names comment is out of date and so I think the confirm would not pass the _ok checks in arm8_configProofScript.sml (also that file needs to be updated).

Here's a names mapping with updated comments that I think should work:

```source can use 29 regs (0-28),
target's r18 should be avoided (platform reserved),
target's r26 must be avoided (extra encoding register),
target's r31 must be avoided (stack pointer),
source 0 must represent r30 (link register),
source 1-4 must be r0-r3 (1st 4 args),
top three (26-28) must be callee-saved (in r19-r29)

0  -> 30
1  -> 0
2  -> 1
3  -> 2
4  -> 3
5  -> 5
6  -> 6
7  -> 7
8  -> 8
9  -> 9
10 -> 10
11 -> 11
12 -> 12
13 -> 13
14 -> 14
15 -> 15
16 -> 16
17 -> 17
18 -> 4
19 -> 19
20 -> 20
21 -> 21
22 -> 22
23 -> 23
24 -> 24
25 -> 25
26 -> 29
27 -> 27
28 -> 28

29 -> 18
30 -> 26
31 -> 31

oskarabrahamsson:
I updated the configs following a discussion on Slack with @tanyongkiam. 

(I ran my tests again with the new config, and the compiler still produces working binaries.)

myreen (approved):

tanyongkiam:
The updated mapping for ARM8 looks good to me (assuming tests pass)

Issue #825, opened 2021-04-25
Negative floating point literals are those of C


konrad-slind:
The use of sscanf in the FFI to handle floating-point literal parsing means that ~42.0 in ML syntax gets translated to the wrong number.

val d = Double.fromString "~42.0";
val _ = print ("~42.0 ----> "^Double.toString d^"\n");

val d = Double.fromString "-42.0";
val _ = print ("-42.0 ----> "^Double.toString d^"\n");

HeikoBecker:
Would you like  `Double.fromString` to  treat `~` like `-`?

konrad-slind:
I think so. Should be like integer literal syntax. I noted in the Slack
chat that one would like

   ~(fromString s) = fromString (~s)

to hold.

myreen:
Closed by PR #826

PR #826, opened 2021-04-27
Extend Double.fromString with a preprocessing step

HeikoBecker:
As discussed in the CakeML slack and in #825 , `Double.fromString` was not properly handling the `~` in its input. This PR makes both `-` and `~` act like unary negation for the `fromString` function by adding a simple preprocessing function.

### DIFF
--- a/basis/DoubleProgScript.sml
+++ b/basis/DoubleProgScript.sml
@@ -26,11 +26,18 @@ val _ = ml_prog_update open_local_block;
 
 val _ = ml_prog_update open_local_in_block;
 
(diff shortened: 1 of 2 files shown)


konrad-slind:
This looks like the right thing. However, note that isPrefix "~" " ~42.0" returns false (note leading space). But (at least in PolyML), Real.fromString " ~42.0" succeeds. This is an annoying detail, I admit.  We could simply ban leading whitespace in the argument  to fromString, or add some code to jump past it.

HeikoBecker:
I have dropped the `isPrefix` check in the latest commit.

PR #827, opened 2021-05-03
Implement parsing of hex integer literals

myreen:
With this addition, one can write programs with integers literal expressed in hex. The following example compiles and prints `255` when run.

    val _ = print_int 0xFF;


mn200:
Awesome; thanks for getting around to this!

PR #833, opened 2021-09-12
Improve compiler error messages

myreen:
The new parse error messages look as follows:
```
$ echo "val n = 5; val val" | ./cake
### ERROR: parse error
Parsing failed at line 1

val n = 5; val val
           ^^^

Program exited with nonzero exit code.
```

Similarly, errors in type inference look as follows:
```
$ echo "val n = 5    5 " | ./cake
### ERROR: type error
Type mismatch between int and int -> _0 at line 1

val n = 5    5
        ^^^^^^

Program exited with nonzero exit code.
```

### DIFF
--- a/basis/basis_ffi.c
+++ b/basis/basis_ffi.c
@@ -158,16 +158,17 @@ int hasT = 0;
 void cml_exit(int arg) {
 
   #ifdef STDERR_MEM_EXHAUST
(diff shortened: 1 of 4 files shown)


mn200 (approved):
Looks great!

konrad-slind:
I agree

PR #834, opened 2021-09-16
Add a CEK-style semantics for CakeML declarations and prove equivalence

hrutvik:
Also re-sync HOL/Lem in `semantics/alt_semantics`


myreen (approved):
Very good! And you even updated comments and READMEs.

PR #843, opened 2021-11-08
Ignore escaped newlines in string literals

### LINKED ISSUE #830: Compiler includes linebreaks inside string literals when computing locations for error messages
When reporting line numbers in errors, the parser--or possibly the machinery surrounding it--includes linebreaks found inside string literals. 

Here is an example file with an error on line 3:
```sml
(* foo.cml *)
val _ = "\n\n\n\n";
val _ = this_causes_an_error "foo";
```
But the compiler concerns line 7:
```
~ cake < foo.cml
### ERROR: type error
Undefined variable: this_causes_an_error at location starting at row 7 column 9, ending at row 7 column 28
```

This error is present in the latest version of the compiler currently available: https://github.com/CakeML/cakeml/releases/tag/v1535.

oskarabrahamsson:
Closes #830.

### DIFF
--- a/semantics/lexer_funScript.sml
+++ b/semantics/lexer_funScript.sml
@@ -75,7 +75,7 @@ Definition read_string_def:
       case TL str of
       | #"\\"::cs => read_string cs (s ++ "\\") (next_loc 2 loc)
       | #"\""::cs => read_string cs (s ++ "\"") (next_loc 2 loc)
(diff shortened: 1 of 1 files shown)


mn200:
Won't you need to also change `lexer_impl` (in `compiler/parsing`)?

oskarabrahamsson:
It doesn't seem like `lexer_imp` redefines the `read_string` function

mn200:
Ah; cool!

PR #846, opened 2021-11-18
Pattern aliases

oskarabrahamsson:
This PR extends the CakeML syntax and semantics with pattern aliases. The syntax is *identifier* "as" *pattern*, and the semantics of `X as P` when matched against `v` is to bind `X` to `v` and then match `P` against `v`.

For example, you can now write something like
```sml
case xs of (h as (x,y))::t => ...
```
and it should have the same semantics as if you wrote:
```sml
case xs of h::t =>
  case h of (x,y) => ...
```


talsewell:
Looks good to me.

I'd like to add some meaningful feedback but I can't think of anything else worthwhile to say.

myreen (changes requested):
All the changes look very good. However, I would like to request that you add an example to [how-to.md](https://github.com/CakeML/cakeml/blob/master/how-to.md) documenting/advertising this feature. The code snippets that are indented in that markdown are regression tested. This means that one, for example, can't write `...` in such code snippets.

mn200:
Regression failure is a problem obviously, but it'd also be good if you added test-cases to `compiler/parsing/tests`.

oskarabrahamsson:
I have fixed the regression test failure, but there are more things to fix before this PR can be tested again: I had forgotten about the type system.

oskarabrahamsson:
> Regression failure is a problem obviously, but it'd also be good if you added test-cases to `compiler/parsing/tests`.

I have now added a few tests of the as-pattern to cmlTests, and fixed some of the testing code that resulted in a few failing tests being accepted, because some test results were not checked.

mn200 (approved):

PR #855, opened 2022-01-05
Add a heap to embody misc; use it in semantics

mn200:
As _per_ a conversation with @myreen, show how we might add a base CakeML heap to help with build times.


mn200:
I think I now understand why this is failing.  Basically, `source_evalProof` (and no doubt others) says that it wants its grammar context to be that of `semantics` (more or less), and due to vagaries in the `semantics` directory, `shift_seq` is or is not in that context.  In `master`, `shift_seq` (and all of `misc`'s other constants) is in `lexer_fun`'s grammar, and so it is in `semantics` also.  In this PR's branch, `lexer_fun` doesn't have it, so `semantics` doesn't have it either.

mn200:
If more of the script files in `semantics` obeyed the rule that everything should be opening `preamble`, this problem would go away, but of the 13 script files there, only 4 do this.  So, the easy fix is to fix all the files in `semantics` to obey our own rules. 

But this is gross in other ways: do you really want all of `misc`'s cruft in semantics?  I'd be more inclined to fix `source_eval` to make its dependency on `misc`'s `shift_seq` more explicit.

PR #856, opened 2022-01-13
Printing

talsewell:
Printing extensions. Various mechanisms for defining functions
named pp_X for each type name X.


myreen:
This will be merged as part of an upcoming PR for the REPL.

PR #857, opened 2022-01-15
Regression test for eval-bitmaps branch

myreen:


myreen:
This regression test is no longer needed.

PR #860, opened 2022-02-06
A verified cyclicity checker, for candle with overloading

arolle:
This PR further extends `candle/overloading`.
We add a verified cyclicity checker algorithm that can check termination of dependencies that arise from definitions, especially of theories of higher-order logic with (ad-hoc) overloading.
We slightly revise some of the earlier added theory in `syntax/`, and synthesise a checker (with an unverified parser) in `ml_checker/`.
Further, `ml_kernel/` contains a verified theorem prover kernel for HOL with overloading, that makes use of the cyclicity checker (to replace a non-termination assumption).


myreen:
As far as I can see, this PR adds three new directories to the repo:

myreen:
Please also update this [README.md](candle/overloading/README.md) by doing a `Holmake README.md` in `candle/overloading`.

arolle:
@myreen , I've resolved your comments in the recent two commits.

myreen:
The new directories built successfully. I'll merge this PR.

PR #861, opened 2022-02-08
Pancake proof update for BaseAddr and some more

mktnk3:
- give BaseAddr semantics and update the proofs
- update the proofs for "first_name" offset change
- update time_to_pan compiler proofs for BaseAddr change

Against `pan-to-mc`, this PR contains the diffs from the rebase.


IlmariReissumies (approved):
Looks great!

I think this is ready to merge. Tests fail, but regressions tests log suggest that everything `pancake` builds successfully, and the failures are the same as those from master, and occur in directories that do not depend on anything that's changed here.

myreen (approved):

PR #866, opened 2022-03-10
Express `semantics_prog` in terms of big-/small-step semantics

hrutvik:
Builds on #834 to prove theorems expressing `semantics_prog` in terms of the relational big-step and CEK small-step semantics in `semantics/alt_semantics`. Also attempts to re-organise `semantics/alt_semantics/proofs` a bit, and removes a stray Lem file.


myreen (approved):
Excellent!

PR #892, opened 2022-06-22
Fix double type printing and basis code generation

HeikoBecker:
With the new Icing tools and value trees, the type for double arithmetic is not word64 anymore. This was not reflected in the type inference pretty-printing and the basis functions. This PR changes this such that `Double.double` is a printed type and the basis functions in `DoubleProg` are properly generated


myreen:
Closes #893

PR #903, opened 2022-08-16
Candle REPL fixes

oskarabrahamsson:
Fixes the following issues with the Candle REPL:
- The lexer now leaves type variables untouched. Previously, writing a type variable could get the lexer stuck scanning an infinitely long character literal. (The Caml syntax uses `'`-ticks for character literals.)
- Change the syntax of the loading directives to match exactly that used in the HOL Light sources. The new directives are:
  * `loads "file";;` to load 'file' and add it to a list of loaded files;
  * `needs "file";;` to load 'file' if it wasn't previously loaded, and add it to a list of loaded files;
  * `#use "file";;` to load a file without adding it to the list of loaded files.
- Add a stub for pointer equality (`(==): 'a -> 'a -> bool`) that always evaluates to false.


oskarabrahamsson:
Err, looks like these commits got tacked onto the  candle-compute-v3 branch. Please ignore.

PR #905, opened 2022-08-18
Add Word8.toIntSigned and Word64.toIntSigned

### LINKED ISSUE #896: Signed Word to Int
(Feature request)

We already have Word64.toInt that treats the word as an unsigned integer. It would be useful to also have a version that treats the word as a signed two's complement integer.

myreen:
Closes #896


myreen:
The regressions test failed due to some unreliability of the regression worker (oven2), but I believe the change should be fine.

PR #917, opened 2022-11-07
Retire Env_id

talsewell:
On writing the Eval paper, it becomes clear that Env_id is an extra feature that isn't actually needed anywhere.

No particular urgency to remove it, but I feel more comfortable not mentioning Env_id in the paper if it's on the way out.


talsewell:
Oops, I've realised something. Don't merge this PR.

PR #924, opened 2023-01-25
Pancake compiler translation

Remyjck:
Changes include:

- Integrating a front-end for the pancake compiler  into the cakeML compiler in compiler/compilerScript.sml
- Adding the translation files pancake_lex, pancake_parse, from_pancake64 and from_pancake32
- Fixing an issue with error propagation through changing locations in pancake/panLexerScript.sml
- Removed some mutual recursion from the prog datatype in pancake/panLangScript.sml (and propagated these changes)


IlmariReissumies:
Thanks for submitting! Closing this as it's part of #927.

PR #931, opened 2023-02-23
Change name of Candle type cexp to cval

myreen:


myreen:
This will be merged as part of another PR.

PR #954, opened 2023-05-15
Added Pancake syntax for function declarations

Tiana-J-TU:
these changes to the Pancake parser allow for writing Pancake functions in concrete syntax


IlmariReissumies:
Thanks for doing this!

PR #958, opened 2023-06-28
Pancake to target correctness proof (resource aware version)

mktnk3:
Update the pan_to_target compiler correctness proof with `extend_with_resource_limit’` and the explicit condition for the equality.


IlmariReissumies (approved):
Nice!

myreen (approved):

PR #976, opened 2023-12-11
Make ffiname a sum of SharedMem and ExtCall

Gordon-Sau:
This PR adds `SharedMem` to ffiname. ffiname can now either be a shared memory operation (`SharedMem`) or a string that represents the name of the foreign function (`ExtCall`).


mn200 (approved):
Looks good to me!

PR #988, opened 2024-03-19
Pancake: concrete syntax for shared memory

mktnk3:
This PR introduces the concrete syntax for Pancake shared memory access.
The details of the changes are as follows:

- syntax for load/store (word/byte access) are changed to accomodate the shared memory instructions

strb  ->  st8   (for both internal/shared memory)
str     ->  stw   (for both internal/shared memory)
ldb     ->   ld8   (for both internal/shared memory)
lds     ->   lds   (for internal memory only; load shape)
               ldw  (newly added; for word access shared memory)

shared memory instructions are prefixed with '!' (so, we have !ldw, !ld8, !stw, and !st8) 

- ffi names are now prefixed with '@' instead of '#'

- the example file is updated (add a shared memory example, update existing ones for load/store changes)


IlmariReissumies (approved):
Looks good.

PR #990, opened 2024-03-25
Syntax fix for RUP

tanyongkiam:


tanyongkiam:
I have tested locally (the compilation)

PR #991, opened 2024-04-13
Add arm8 bootstrap

LudvikGalois:


basis/basis_ffi.c
```diff
@@ -143,7 +143,7 @@ void ffiopen_out (unsigned char *c, long clen, unsigned char *a, long alen) {
   #ifdef EVAL
   int fd = open((const char *) c, O_RDWR|O_CREAT|O_TRUNC, S_IRUSR|S_IWUSR|S_IRGRP|S_IROTH);
   #else
-  int fd = open((const char *) c, O_RDWR|O_CREAT|O_TRUNC);
+  int fd = open((const char *) c, O_RDWR|O_CREAT|O_TRUNC, 0600);
```
oskarabrahamsson, line 146: I think it would be better to replace this whole chunk with what's on line 144. I checked how `TextIO.openOut` works with Poly/ML and Mlton on my machine, and both seem consistent with this (i.e., -rw-rw-r--).

compiler/bootstrap/compilation/arm8/32/proofs/README.md
```diff
@@ -0,0 +1,5 @@
+This directory contains the end-to-end correctness theorem for the
+32-bit version of the CakeML compiler.
+
+[x64BootstrapProofScript.sml](x64BootstrapProofScript.sml):
```
oskarabrahamsson, line 4: arm8BootstrapProofScript.sml

compiler/bootstrap/compilation/arm8/64/cake-sexpr-64
oskarabrahamsson, line 1: I don't think we want these two checked in

oskarabrahamsson:
* To get an arm64 compiler binary out of the bootstrap runs, you also need to add your target to this file: https://github.com/CakeML/cakeml/blob/master/developers/build-sequence.
* The 32-bit arm8 target seems unnecessary: are there are any armv8 processors that only support aarch32?

LudvikGalois:
> * To get an arm64 compiler binary out of the bootstrap runs, you also need to add your target to this file: https://github.com/CakeML/cakeml/blob/master/developers/build-sequence.

The arm64 bootstrap needs to be run on an arm64 system, so I don't think it makes sense to add it to the build-sequence, since whatever is running the regression tests is an x86-64 system.
 
>     * The 32-bit arm8 target seems unnecessary: are there are any armv8 processors that only support aarch32?

I'll remove the 32-bit target.

tanyongkiam:
> > * To get an arm64 compiler binary out of the bootstrap runs, you also need to add your target to this file: https://github.com/CakeML/cakeml/blob/master/developers/build-sequence.
> 
> The arm64 bootstrap needs to be run on an arm64 system, so I don't think it makes sense to add it to the build-sequence, since whatever is running the regression tests is an x86-64 system.

The in-logic compilation runs inside HOL4 and can produce the arm64 version of the compiler's .S file even while ran on an x86-64 server. It would be good to have it in the regression so that we can distribute the arm64 version of the .S file (like we do for x86-x64).

LudvikGalois:
> The in-logic compilation runs inside HOL4 and can produce the arm64 version of the compiler's .S file even while ran on an x86-64 server. It would be good to have it in the regression so that we can distribute the arm64 version of the .S file (like we do for x86-x64).

Currently the makefile rule for the `cake-arm8-64.tar.gz` target includes a quick test to make sure the bootstrap works (because the x86-64 bootstrap had that test), and that _won't_ run on an x86 machine. I'll remove that test since the x86-32 bootstrap doesn't have it, and then add arm8-64 as a target to the build-sequence file.

Also, sorry this is taking so long; I wanted to re-run the bootstrap after removing the arm8-32 bootstrap to make sure I didn't break anything.

tanyongkiam:
Great, thanks.

Also, you don't need to run this yourself. Once the relevant files are in the build-sequence, open PRs will get picked up by the CakeML regression testing system.

oskarabrahamsson:
Won't this PR need an update now that #1002 has been merged?

myreen:
Closing because this PR has been continued as #1022.

PR #1004, opened 2024-06-25
Added lower ("<+") comparators to Pancake concrete syntax.

DoomishGuy:


myreen:
Could the syntax for less/greater-equal be <=+ and >=+ so that they are consistent with the syntax from wordsTheory?

PR #1016, opened 2024-07-18
pancake: patch crep_to_loop to use static Call

talsewell:
Adjust the crep_to_loop compilation phase to spot the case where the target is a constant "Label X" expression, and produce a call to a known target on the loopLang side. Hopefully this should result in static calls in the final binary, rather than needless indirect calls. Frustratingly, for a simple change, quite a lot of the proof script had to be amended.

I think the tests will pass, but I haven't run them comprehensively on my machine, in particular I haven't thought about
translation.

### DIFF
--- a/compiler/bootstrap/translation/from_pancake32ProgScript.sml
+++ b/compiler/bootstrap/translation/from_pancake32ProgScript.sml
@@ -253,6 +253,8 @@ val _ = translate $ spec32 compile_crepop_def;
 
 val _ = translate $ spec32 compile_exp_def;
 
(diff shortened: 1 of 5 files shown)


IlmariReissumies:
Thanks for making the proofs cleaner while at it---there's way too much 'TRY` spaghetti in the Pancake proofs...

talsewell:
> Thanks for making the proofs cleaner while at it---there's way too much 'TRY` spaghetti in the Pancake proofs...

Well, yes, that's one of the main reasons I had to change so much. I just couldn't figure out how bits of the old proof worked.

PR #1029, opened 2024-07-28
Add 32-bit shared memory reads/writes to Pancake

### LINKED ISSUE #1000: Support half word loading/storing in Pancake
Currently there're ways to load/store words and bytes, but not a direct way to load/store half words.
It'd be a good usability improvements to support those operations, with something like:
```
var half = ldh address;
sth address, half;

!ldh half, address;
!sth half, address;
```

IlmariReissumies:
Closes #1000 (but leaves the issue of 16-bit instructions dangling)


pancake/pan_passesScript.sml
```diff
@@ -312,7 +312,7 @@ Definition crep_prog_to_display_def:
                    | Load32 => [String (strlit "load"); String (strlit "halfword")]
                    | Store => [String (strlit "store"); String (strlit "word")]
                    | Store8 => [String (strlit "store"); String (strlit "byte")]
-                   | Store8 => [String (strlit "store"); String (strlit "halfword")]) in
+                   | Store32 => [String (strlit "store"); String (strlit "halfword")]) in
```
IlmariReissumies, line 315: Thank you!

PR #1048, opened 2024-08-23
Expose main return value when using `--main_return=true`

halogentlepersuasion:
Includes a slight refactor of MEP assembly
Use by replacing `void` return type of `cml_main` in C file


halogentlepersuasion:
Because of details in the generated stubs, this fix doesn't yet work since the return value is overridden by the time the assembly template gets access to it again

PR #1056, opened 2024-09-21
Use new Quote syntax in basis

myreen:


dnezam:
Fully updating to the new Quote syntax depends on https://github.com/HOL-Theorem-Prover/HOL/issues/1313 - in particular, there is at least once instance in `characteristic/examples/cf_examplesScript.sml` that is affected by this.

PR #1072, opened 2024-10-17
Increase Infixl in npbcScript

dnezam:
Hopefully fixes
`GrammarError "Attempt to have differently associated infixes (LEFT and RIGHT) at same level(501)"`.


IlmariReissumies:
@myreen pushed this exact tweak to master in #c7db3a8 so the PR is redundant, thanks though :)

PR #1092, opened 2024-11-13
Remove non-working example

xrchz:


dnezam:
Before merging this PR, we should make sure that we have created issues to restore theories that are of interest.

PR #1099, opened 2024-11-27
Make ml_progLib build `evaluate_dec_list` rather than `evaluate_decs`

myreen:
This PR speeds up bootstrap translation by about 2 hours. This is achieved by pushing some of the work done by `ml_progLib` into the in-logic compilation performed by `cv_compute`.

This PR makes `ml_progLib` build a proof about `evaluate_dec_list` (new here) rather than `evaluate_decs` (official CakeML semantics). These are equivalent if the program is syntactically okay according to a simple syntactic check.


myreen:
This PR also adds a new script to `developers`. By running,

    $ developers/build-everything.py

the script will run `Holmake` in every directory of the build sequence. This is useful for local testing before pushing to a regression worker.

PR #1102, opened 2024-12-07
Remove candle/set-theory/zfc

dnezam:
Seems like this folder has bit-rotted away. The issue
```
HOL_ERR {message = "No consistent parse", origin_function = "typed_parse_in_context", origin_structure = "Parse"}
```
can be fixed by adding `'` as an infix operator (probably left associative, with 2000 precedence), so that the overload and its usages work. However, after that, one gets the error
```
error: Value or constructor (setSpec_grammars) has not been declared
 Found near
   (
      type_grammar.remove_abbreviation (#1 setSpec_grammars) "reln",
      #2 setSpec_grammars
      )
```
if one tries to now build that folder.

At the time of writing, it seems that it was [created ~8.5y ago](https://github.com/CakeML/cakeml/commits/2d4728905a58811ef8a61a764af76c01e02f0812/flame), and [moved to its current home ~7y ago](https://github.com/CakeML/cakeml/commit/003d997302599efed169665d821607809e29d2da) with the comment:
> It looks like this development is in hibernation if not abandoned


xrchz:
Sounds good to me - the actually used set theory formalisation is still there so this can live in git history only.

dnezam:
Does this mean we shouldn't create an issue to restore it? It's hard for me to gauge the interest/value in it.

PR #1119, opened 2025-01-03
Update build-instructions.sh

ordinarymath:
Note that brew polyml is broken and update polyml build instructions to enable using gmp.

### DIFF
--- a/build-instructions.sh
+++ b/build-instructions.sh
@@ -3,15 +3,15 @@
 
 ## build Poly/ML
 ## alternatively on macOS
-# Note that as of 3/1/2025 brew from polyml is broken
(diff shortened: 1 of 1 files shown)


xrchz (approved):

build-instructions.sh
```diff
@@ -3,11 +3,15 @@
 
 ## build Poly/ML
 ## alternatively on macOS
+# Note that as of 3/1/2025 brew from polyml is broken
```
xrchz, line 6: Poly/ML in brew instead of "brew from polyml"?

ordinarymath: Fixed. Thanks.

PR #1129, opened 2025-02-03
Remove pair_case_eq as it is in pairTheory

ordinarymath:


ordinarymath:
There 's also a `bool_case_eq` which is equal `CaseEq "bool"` directly above which should probably be upstreamed.

PR #1141, opened 2025-03-10
Streamline license

dnezam:
Hopefully this will allow GitHub to automatically detect what license we are using.


LICENSE
```diff
@@ -0,0 +1,28 @@
+BSD 3-Clause License
```
xrchz, line 1: I do think the project name should appear somewhere here - can the previous top line also be included somehow, maybe as a second line?

digama0: I don't think that's traditional, license texts are generally kept completely pristine (i.e. `diff` reports no changes) to make it completely obvious that there is no legal issue a la crayon licenses. The name of the project is clearly described in the readme. The title of this document is an identifier of the license itself, and unless you are modifying and rebranding it there is no need to have a "CakeML License" header. [Example](https://github.com/rust-lang/cargo/blob/master/LICENSE-APACHE)

PR #1142, opened 2025-03-12
Remove more misc

ordinarymath:


xrchz (approved):

ordinarymath:
This is more troublesome breakage the issue is that `bool_case_eq` and `CaseEq "bool"` are different

PR #1146, opened 2025-03-24
Fix Pancake static checker warnings causing non-zero exit code during compilation 

halogentlepersuasion:
The current mechanism for determining the exit code at the end of compilation checks if the `stderr` output starts with `###`, which is printed at the beginning of error messages. Warnings, as generated by the static checker, mistakenly used this prefix, causing otherwise successful compilation to be treated as failure in eg. build scripts. This PR simply removes this prefix match.


ordinarymath:
@halogentlepersuasion slightly off-top but is the readme generated for pancake/static_checker/panStaticExamplesScript.sml intended see #1145

halogentlepersuasion:
@ordinarymath Yes, it is intended

PR #1165, opened 2025-04-30
Pancake internal 32-bit load/store

mktnk3:


ordinarymath:
I think that my changes to stackProps in #1166 will conflict will fix them

PR #1177, opened 2025-05-12
Clos do word app

ordinarymath:
This is a refactor the do_app for languages that use closLang$op for the word case.


tanyongkiam:
Tested locally.

PR #1195, opened 2025-07-17
Update PB for dominance-with-auxliaries

tanyongkiam:


tanyongkiam:
Tested on oven2 against HOL `a81865b9bac5e3018a070980ce8f4a392c66c9c6`

Issue #979, opened 2024-02-04
Replace standard list sort with better sort


tanyongkiam:
The default list sort has quadratic behavior on sorted (or nearly sorted) lists because it picks the pivot from the head of the list.

This issue is about potentially replacing:

https://github.com/CakeML/cakeml/blob/alist_subst/basis/ListProgScript.sml#L379

with the "OCaml-inspired" mergesort available in HOL.

tanyongkiam:
#1206 partially resolved this by defining a central "sort" that is shared across the codebase.

The next step is to implement a fast sort replacing this standardized sort.

tanyongkiam:
Just to add to the above, ideally, the use of arrays should be *internal* to the list version, i.e., the final spec for the list-based sort is still compatible with the translator.

In this way, a lot of code in CakeML would automatically benefit from the fast array-based sort (internally).

tanyongkiam:
Yes, it is, see e.g.,: https://github.com/CakeML/cakeml/blob/master/compiler/backend/reg_alloc/reg_allocScript.sml#L1457

However, I would have thought that this could be relatively painless within CF as well.

PR #1206, opened 2025-08-08
Define mllist$sort and use it 

tanyongkiam:
Resolves issue #979. Potential follow up is to improve `mllist$sort`


tanyongkiam:
I confirmed that this builds on oven2

PR #1210, opened 2025-08-14
Revised floats

mn200:
I am pretty confident this is building against `master` but hope that making the PR will prompt the regression machinery to check.  

The essence of the change is the removal of real numbers, fpOptimise and fp/real-trees from the semantics.  There are knock-on effects down to `clos` because `flat` uses the semantics operations and those have changed.  `clos` keeps its existing operations, so nothing below there needs changing.


mn200:
I'm working on getting the merge to work.

PR #1215, opened 2025-08-16
suggested fix for blocktypes

CHRChhak:
Prior to this PR, semantics of `Blocks` (& Loop) expected block types to be function types (:= valtype list # valtype list). For example, the semantics checked to make sure the stack had enough "arguments" (given by the function type) to the block. 

This PR suggests simplifications that in line with having block types be result types (:= valtype option)

### DIFF
--- a/compiler/wasm/wasmSemScript.sml
+++ b/compiler/wasm/wasmSemScript.sml
@@ -40,10 +40,16 @@ Datatype: result
   | RTimeout
 End
 
-(* Returns the function type for tb *)
-Definition functype_of_blocktype_def:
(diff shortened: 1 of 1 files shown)


tanyongkiam:
I discussed this with @wildptr and we decided to merge it.

Note for future reference: this makes the semantics of Block and Loop closer to Wasm 1.0 (which does not support arbitrary function types for block types).

If more flexibility is required for block types in the future, then we may have to restore the old semantics.

CHRChhak:
great @ merge

re future dev: But it should be added on top (and not replace the current semantics) right?

PR #1252, opened 2025-10-23
First pass at speeding up CF automation

tanyongkiam:
Work done by @dnezam


ordinarymath:
I think it was 30 mins roughly before I did the lift out ``REWR_CONV _`` calls in the arith decision procedures and it got to slightly more than 20 mins.

PR #1253, opened 2025-10-24
Add basic profiling support for generating flame graphs for theories and proofs

dnezam:
I tried to document thoroughly, so I hope it's not hard to use.

For example, in `basis/HashtableProofScript.sml`, adding
- `Profiler[qualified] profiler_base profiler_cfml` to the `Libs` section
- `val _ = Profiler.export_cwd "HashtableProof.txt";` to the end of the file

and adding `$(CAKEMLDIR)/profiler` to `INCLUDES` in `basis/Holmakefile`, it is possible to use [FlameGraph](https://github.com/brendangregg/FlameGraph) like this:

```
flamegraph.pl --flamechart HashtableProof.txt > HashtableProof.svg
```

to generate the following flame chart:
HashtableProof
The actual `svg` is interactive and allows us, for example, to zoom onto theorems and also check how much time they took in milliseconds and percentage.

### DIFF
--- a/developers/build-sequence
+++ b/developers/build-sequence
@@ -29,6 +29,9 @@ characteristic
 translator/monadic
 translator/monadic/monad_base
 
+# profiler
+profiler
+
 # basis library
 basis
 


ordinarymath:
To limit what's exported as a cleaner api.

PR #1258, opened 2025-11-08
Another pass at speeding up CFCML

dnezam:
The optimizations were mainly guided by looking at hotspots in `npbc_arrayProg` using the profiler. The speedup for it seems to be 10 minutes (32m => 22m). The proofs now take around 10 minutes; the rest is other stuff (probably translation; maybe that should also be profiled in the future).

Also, `TextIOProof` now takes a bit over 7 minutes.

### DIFF
--- a/characteristic/cfTacticsLib.sml
+++ b/characteristic/cfTacticsLib.sml
@@ -108,9 +108,9 @@ val reduce_conv =
     (DEPTH_CONV (
       List.foldl (fn (pat, conv) => (eval_pat pat) ORELSEC conv)
                  ALL_CONV reducible_pats))
-    (* It seems that the next line makes around 29s faster; it would be good
-       to confirm this, though. Maybe doing basic simplifications first results
-       in a smaller term, meaning less work is done in the following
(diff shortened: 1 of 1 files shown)


characteristic/cfTacticsLib.sml
```diff
@@ -13,27 +13,23 @@ open xcf;
...
-  ("extend_env_rec_ellipsis", ``extend_env_rec _ _ _ _ _``, ellipsis_pp),
-  ("extend_env_with_ellipsis", ``extend_env _ _ _ with v := _``, ellipsis_pp),
-  ("extend_env_rec_with_ellipsis", ``extend_env_rec _ _ _ _ _ with v := _``,
-   ellipsis_pp)
-]
-
-fun hide_environments b =
+  val ellipsis_pp = constant_printer "(…)"
+  val printers = [
+    ("extend_env_ellipsis", ``extend_env _ _ _``, ellipsis_pp),
```
ordinarymath: I don't think you should inline calls to Term.parse

dnezam: Good point.

PR #1278, opened 2025-12-05
Try to store only delta's in translator Part 1

ordinarymath:
This is quite a bad way but still better than
the old way
Progress list.

1. unpack_types --done
2. unpack_v_thms
3. unpack_cons_names
4. unpack_type_mods


ordinarymath:
DO NOT MERGE. I'm only opening a PR to test with regression

Issue #1271, opened 2025-11-24
Tidy up and add primitives to source AST op


myreen:
I recently made a PR (https://github.com/CakeML/cakeml/pull/1270) that added `Test` to source `op` datatype. This issue is about doing more of these kinds of rearranging and additions to source `op`.

My PR added:
```
   | Test test prim_type
```
where
```
Datatype:
  test = Equal | Less | Less_alt | LessEq | LessEq_alt
End
```
and
```
Datatype:
  prim_type = BoolT
            | IntT
            | CharT
            | StrT
            | WordT word_size
            | Float64T
End
```
The plan is to also add a primitive that collects all direct conversions between primitive types:
```
   | FromTo prim_type prim_type
```
and also:
```
   | Arith arith_op prim_type
```
where arith will collect `+`, `-`, `*`, `div`, `mod`, `and`, `or`, `xor`, `neg`, `not`, `fma`.

It would also be good to structure `test` better. Something like:
```
  test = Equal | Cmp cmp | AltCmp cmp
```
where cmp is `<`, `<=`, `>=`, `>` and `AltCmp` is signed compare for words and fast string comparison for strings (https://github.com/CakeML/cakeml/issues/1160).

The intention is not to give every possible combination meanings (e.g. what's mod of two strings?). The structure of the definition is partly motivated by streamlining type inference and also it makes it obvious which combinations exist (rather than the ad hoc naming and duplication we have currently, e.g. `Less`, `Lt`, `Less_alt` and `Plus`, `Add`, `FP_Add`).

This redesign is consistent with slightly more exotic operations, e.g. doing bitwise-operations on arbitrary precision integers (https://github.com/CakeML/cakeml/issues/1074) or doing bitwise operations on two strings.

xrchz:
Next step for step 1: Make `supported_arith` include all word ops and float ops.
Then do step 2: remove the old ops from the AST.

PR #1296, opened 2026-01-05
Refactor primitive ops

xrchz:
Work on #1271

### DIFF
--- a/compiler/repl/evaluate_initScript.sml
+++ b/compiler/repl/evaluate_initScript.sml
@@ -640,6 +640,7 @@ Proof
          store_lookup_def, copy_array_def, store_assign_def,
          CaseEq"sum"]
     \\ Cases_on ‘a’ \\ Cases_on ‘ty’
+    \\ TRY(rename1 ‘WordT w’ \\ Cases_on ‘w’)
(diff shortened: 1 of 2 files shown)


dnezam:
Seems like it is using `TRY` which isn't ideal (but I guess somewhat understandable in that situation), but it also looks like it is relying on auto-generated variables names, which should be fixed if possible

compiler/backend/proofs/flat_to_closProofScript.sml
```diff
@@ -1311,20 +1329,24 @@ Proof
   \\ fs [flatSemTheory.do_app_def,list_case_eq,CaseEq "flatSem$v",PULL_EXISTS,
          CaseEq "ast$lit",store_assign_def,option_case_eq]
   \\ gvs [AllCaseEqs()]
-  \\ Cases_on `ty`
-  (* Eliminate impossible types: BoolT, CharT, StrT, WordT all make do_arith return NONE *)
+  \\ Cases_on `ty` \\ TRY (Cases_on ‘w’)
```
dnezam: This (and other TRY (Cases_on `w`)) seems to rely on an auto-generated name

PR #1298, opened 2026-01-08
Use mlstring in ast and related theories

### LINKED ISSUE #1203: Replace `string` by `mlstring` in the CakeML source AST
We should try to use the translator's brittle "treat `string` as if it is `mlstring`" feature much less. 

This issue is about replacing all occurrences of `string` in [semantics/astScript.sml](https://github.com/CakeML/cakeml/blob/master/semantics/astScript.sml) with `mlstring` and updating the CakeML codebase for this change.

dnezam:
Additionally, we made the following changes:

*  Move typeDecToPP

Avoids basis/pure from depending on semantics

*  Various additions to mlstringLib and mlstringSyntax

add_mlstring_compset (mlstringLib), mk_mlstring and dest_mlstring
(mlstringSyntax)

* Update comment describing IO_fs in fsFFI

Based on VSTTE18 paper

* Flip equality in INSTREAM_LINES + Make lines_of_gen_nil a simp

The original direction of the equality could lead to an assumption of the form
`[] = lines_of_gen ...` to be generated, which in combination with a call to the
simplifier would wreak havoc in the environment in some cases. In particular,
xlet_auto and xapp would not terminate (most likely due to something concerning
eval in reduce_tac).

* Remove unused basisComputeLib

* Delete all uses of use_string_type in translation

Closes #1203

### DIFF
--- a/compiler/bootstrap/compilation/ag32/32/proofs/ag32BootstrapProofScript.sml
+++ b/compiler/bootstrap/compilation/ag32/32/proofs/ag32BootstrapProofScript.sml
@@ -174,8 +174,7 @@ Proof
   \\ disch_then drule
   \\ strip_tac
   \\ simp[ag32_memoryTheory.ffi_exitpcs_def]
-  \\ conj_tac >- (simp[LENGTH_code] \\ EVAL_TAC)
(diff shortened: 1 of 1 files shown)


tanyongkiam:
I am running the regression manually.

PR #1346, opened 2026-03-03
Monadic translator lib fixes

talsewell:
Some fixes to the monadic translator libs to be a little less fragile.

It seems that the examples still run and some corner cases are fixed, to my initial test. Let's see what the regression thinks.


tanyongkiam:
Is this superseded by https://github.com/CakeML/cakeml/pull/1348 ?

talsewell:
Yeah, whatever is broken also affects #1348 , so I'll try to fix them together.

PR #1365, opened 2026-03-25
Switch to more efficient sexp datatype

myreen:


dnezam:
In addition to the proof failure in `sexp_parserProg`, it also swapped out Unicode double quotes with string quotes in `dafny_compilerProg`

tanyongkiam:
the sexp_parser failure is known, I just didn't want to fix it manually yet...

dnezam:
Claude did some really bad merging...

dnezam:
Seems like it builds now. Before this gets merged, the changes should be documented (including checking for regressions).

One example of a silent regression:
```
- if s = "Vsubunsafe" then SOME Vsub_unsafe else
+ if a = «Vsub_unsafe» then SOME Vsub_unsafe else
```

dnezam:
I just briefly looked over it (so there might be more), and want to note:

Is [astToSexprLib.sml](https://github.com/CakeML/cakeml/pull/1365/changes#diff-7d87a7fbe483721a00aa736cf2d06c7d0c0f2fd8d73a1f3e3442e4950801e14e) really necessary, or is it possible to use cv_eval somehow?

The definition of `odestSXNUM_def` is a bit strange:
```
Definition odestSXNUM_def[simp]:
  (odestSXNUM (Atom s) =
    (case fromString s of
     | SOME i => if 0 ≤ i ∧ toString i = s then SOME (Num i) else NONE
     | NONE => NONE)) ∧
  (odestSXNUM (Expr _) = NONE)
End
```
Note how it converts a string it read into an integer and checks for equality. I am guessing its because some canonicity property is required somewhere; however, this seems like a really bad way of achieving that. It might even be a performance regression for S-expressions that might have massive numbers, which I don't think is too unreasonable considering that CakeML supports bignums and rationals.

I think it would be better if this refactor was done much more manually.

PR #1390, opened 2026-05-10
Fix a modelling mistake

tanyongkiam:


tanyongkiam:
I've tested this locally (it only changes an example file without downstream dependencies).

PR #1397, opened 2026-05-22
perf-record --call-graph support (x64-only)

dnezam:
Passing `--perf_callgraph=T` to the CakeML compiler (x64-only) generates an unverified binary that can be profiled with `perf record --call-graph fp`.
The in-logic compiler uses the default config in `x64_config`.
Thus, to profile binaries such as checkers, the Dafny compiler, the Scheme compiler, etc., it is necessary to set `perf_calls:=T` in the configuration ultimately used by the compiler.


myreen (approved):
Looks good to me

PR #1402, opened 2026-05-31
globally increase num arith cache params 100x

tanyongkiam:


tanyongkiam:
P.S.: this is already tested locally, I just want a regression to run on it to see the numbers.

PR #1407, opened 2026-06-07
First-order version of `peg_exec` and its translation to cv_compute

mzh-100:
This PR introduces a first-order version of `peg_exec` in file `compiler/parsing/pegexec_cml_foScript.sml`, followed by the translation for use with `cv_compute`.

 ## Current Status & Blockers
I attempted to run benchmarks integrated with `xlrup`, but encountered a translation failure in `examples/xlrup_checker/pegexec_foScript.sml`. Specifically, `cv_translate` fails when processing `fromChars_unsafe`.

As an alternative workaround to evaluate execution times, I have attached a customized benchmark suite in `compiler/parsing/pegexec_cml_foScript.sml`.

Apologies for pushing other unrelated `*.txt` files.


dnezam:
Thanks for opening a PR! It would be good if you could push a commit that removes all the .txt files :)

mzh-100:
> Thanks for opening a PR! It would be good if you could push a commit that removes all the .txt files :)

Hi there, I have pushed a clean version.

tanyongkiam:
Could you clean up the files somewhat? For example, the copy in `examples/xlrup_checker` seems like a misplaced copy.

In addition, I don't think we really need the benchmarking scripts included in the file.

mzh-100:
> Could you clean up the files somewhat? For example, the copy in `examples/xlrup_checker` seems like a misplaced copy.
> 
> In addition, I don't think we really need the benchmarking scripts included in the file.

Hi, I have made the update accordingly.

mn200:
Looks good to me.

PR #1415, opened 2026-06-16
Speed up Candle by 1.88x

dnezam:
On `master`, Candle takes 56m49.502s to run [make_complex.ml](https://github.com/CakeML/candle/blob/master/Multivariate/make_complex.ml) (with `loadt` replaced by `loads`, and the database elements commented out).

With this PR, the same file takes 30m10.587s. The 1.88x comes from calculating 56.5/30 = 1.883.

For reference, HOL Light takes ~37m on my machine, so on `make_complex.ml` Candle is 37/30=1.23x faster.

The main speedup comes from two changes:
1. Use safe_mk_eq instead of mk_eq
2. Do not recreate "constant" types such as `bool` or `aty`

These were discovered using the recently added (unverified) perf call-graph support. The main issue is that `mk_eq` would walk an association list (alist) basically to the end to find the definition of `=`, which I think gets worse the more definitions are made.

Something similar applies to creating types, where one needs to do an alist lookup to find the appropriate type. Additionally, since `bool` and `aty` are non-trivial constructors, they would be reallocated over and over again, wasting memory and (I suspect) breaking pointer equality.

I've also tweaked some other, more minor points. For example:
- I replaced `mk_comb` with `Comb` in `new_basic_type_definition_def`, which matches HOL Light more closely
- I replaced expressions of the form `compare x y = EQUAL` with `x = y` to avoid unnecessarily computing a (ternary) order when we only care about equality 
- I've also made minor adjustments to `alphavars` and `term_union`, which don't make to seem much difference in terms of performance, but brings Candle a tiny bit closer to HOL Light.

Future work:
- Go through the Candle and HOL Light kernel carefully and check whether there are any more performance relevant differences (for example more places where `mk_comb` could be replaced by `Comb`)
- Pointer equality: on a branch where we hacked in pointer equality (both into the kernel and into the source files), according to my notes, make_complex.ml ran in ~26m 12s (3 measurements), which roughly corresponds to a potential 1.15x speed up to be gained.
  - I think most of this time comes from savings in `alphavars`. 
- In `vsubst_aux`, a fast path is commented out with the reasoning that "it doesn't seem to fit Harrison's formalisation". Looking at the perf data for the fastest Candle version (with pointer equality, unverified), it seems that a few percent points are actually spent in INST, which appears to call `vsubst` multiple times. According to an LLM, the argument is basically that if you know that the for `s' = vsubst_aux ilist' s`, s' and s are equal, then by virtue of `ilist' = FILTER (\(t,x). x <> v) ilist` the condition in the branch (`EXISTS (\(t,x). vfree_in v t /\ vfree_in x s) ilist'`) does not hold, and we fall through to `Abs v s'`. However, since `s' = s`, this is the same as the original term `Abs v s`. As mentioned, I believe the perf data supports that this optimization is relevant.

### DIFF
--- a/candle/prover/candle_kernel_permsScript.sml
+++ b/candle/prover/candle_kernel_permsScript.sml
@@ -1299,6 +1299,20 @@ Proof
          ml_translatorTheory.INT_def, perms_ok_def]
 QED
 
+Theorem perms_ok_bool_ty_v[simp]:
+  perms_ok kernel_perms bool_ty_v
+Proof
+  assume_tac $ (bool_ty_v_thm |> CONV_RULE EVAL) >> fs []
+  >> rw [perms_ok_def]
+QED
+
+Theorem perms_ok_aty_v[simp]:
+  perms_ok kernel_perms aty_v
+Proof
+  assume_tac $ (aty_v_thm |> CONV_RULE EVAL) >> fs []
+  >> rw [perms_ok_def]
+QED
+
 (* Functions translated with 'm_translate' should be proved for kernel_perms *)
 
 Theorem perms_ok_the_type_constants[simp]:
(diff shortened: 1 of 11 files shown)


tanyongkiam:
running locally

tanyongkiam:
tested locally

Issue #1425, opened 2026-06-24
Super-linear (O(n^2)) behavior in basis: TextIO.output and String.concatWith


tanyongkiam:
Two basis functions build strings by **repeated copy** instead of the accumulate-then-`concat`-once idiom, giving O(n²) behavior on large inputs. Both share the same root cause, and the basis already demonstrates the correct fix elsewhere (`String.tokens`/`String.fields` were rewritten to substring-based `_alt` forms precisely to avoid this class; `TextIO.inputAll` uses accumulate-then-`concat`-once).

Because CakeML strings are immutable flat byte vectors, `String.substring`/`String.extract`/`^`/`String.concat` all **allocate and copy** their result — there is no slicing/sharing. So copying a growing prefix/suffix once per loop iteration is quadratic.

## 1. `TextIO.output` — O(n²) in the length of the string

`basis/TextIOProgScript.sml` (the `output` definition):

```sml
fun output fd s =
  if s = "" then () else
  let val z = String.size s
      val n = if z <= 2048 then z else 2048
      val fl = Word8Array.copyVec s 0 n iobuff 4
      val a = write (get_out fd) n 0 in
         output fd (String.substring s n (z-n))
  end;
```

Each recursive step calls `String.substring s n (z-n)`, which **allocates and copies the entire remaining suffix** (`z - n` bytes) just to drop the 2048-byte prefix it already wrote. For an input of length `L` with chunk `C = 2048`:

```
Σ_{j=1}^{L/C} (L - C·j) ≈ L²/(2C)
```

so ~`L²/4096` bytes copied — quadratic — against only `Θ(L)` of useful work (each byte is written once). Concretely: ~128 MB copied for a 1 MB string, ~32 GB for 16 MB.

(There is a secondary cost: each `String.substring` lowers — via `CopyStrStr` → `CopyByte T` → `bvl_to_bvi`'s `RefByte`+`CopyByte F` expansion — to a fresh byte-array **allocation that is zero-filled by `Replicate` and then immediately overwritten by the copy**, so each chunk's suffix is effectively touched twice. This is the inefficiency already flagged by the `TODO` at `compiler/backend/bvl_to_bviScript.sml` on the `CopyByte T` case.)

### Fix
`Word8Array.copyVec` already takes a source offset, so thread an index into the original `s` and never re-slice:

```sml
fun output fd s =
  let val z = String.size s
      fun loop i =
        if i >= z then () else
[...]

PR #1426, opened 2026-06-25
Make TextIO.output linear in the size of the string (previously quadratic)

dnezam:
Instead of `TextIO.output` copying the string a quadratic number of times (yikes!), we keep track
of a "pointer" into the string, which should make it linear (and thus faster).

I've also added a new simp (get_mode_fsupdate), which may break proofs outside the CakeML repo (assuming they depend on it).

Progress on #1425


dnezam:
I think this should be squash-merged.

tanyongkiam:
tested locally by @dnezam

PR #1427, opened 2026-06-25
Holprojs

mn200:
Use Holmake-compatible `holproject.toml` files across the repo to remove the need for `INCLUDES` lines and other "junk".


mn200:
Green!

PR #1429, opened 2026-06-29
Refactor usages of state-exception monad

dnezam:
- use `monadsyntax.temp_enable_monad` instead of locals overloads => uses state-exception's ignore_bind (instead of overloading using st_ex_bind)
- use state-exception monad in inferScript.sml (replaces its own definition of a state-exception monad)


tanyongkiam:
Locally tested by @dnezam

PR #1430, opened 2026-07-02
Improve implementation of String.concatWith

### LINKED ISSUE #1425: Super-linear (O(n^2)) behavior in basis: TextIO.output and String.concatWith

dnezam:
- Add List.intersperse to basis
- Fix String.concatWith to not be quadratic in the worst case
- Remove of unused CONCAT_WITH (misc) and concatWith_CONCAT_WITH (mlstring)

Closes #1425

### DIFF
--- a/basis/ListProgScript.sml
+++ b/basis/ListProgScript.sml
@@ -86,6 +86,9 @@ val result = translate list_compare_def;
 val result = next_ml_names := ["concat"];
 val result = translate FLAT;
 
+val result = next_ml_names := ["intersperse"];
+val result = translate mllistTheory.intersperse_def;
+
 (* the let is introduced to produce slight better code (smaller stack frames) *)
 Theorem MAP_let[local]:
     MAP f xs =
--- a/basis/StringProgScript.sml
+++ b/basis/StringProgScript.sml
@@ -46,10 +46,6 @@ val extract_side_thm = Q.prove(
   `!s i opt. extract_side s i opt`,
   rw [extract_side_def, arithmeticTheory.MIN_DEF] ) |> update_precondition
 
-val _ = ml_prog_update open_local_block;
-val res = translate concatWith_aux_def;
-val _ = ml_prog_update open_local_in_block;
-
 val _ = next_ml_names := ["concatWith"];
 val result = translate concatWith_def;
 
--- a/basis/pure/mllistScript.sml
+++ b/basis/pure/mllistScript.sml
@@ -8,6 +8,12 @@ Ancestors
   indexedLists[qualified] toto[qualified]
   sorting mergesort
 
+Definition intersperse_def:
+  intersperse _ []        = [] ∧
+  intersperse _ [x]       = [x] ∧
+  intersperse sep (x::xs) = x::sep::intersperse sep xs
+End
+
 (* ===== TODO: TO BE PORTED TO HOL (better theorems for mergesort_tail) ===== *)
 Theorem merge_tail_MEM:
   !negate R xs ys acc. MEM x (merge_tail negate R xs ys acc) = ((MEM x xs) \/ (MEM x ys) \/ (MEM x acc))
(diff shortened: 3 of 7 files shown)


tanyongkiam:
tested locally

PR #1431, opened 2026-07-04
Some stashed PB checker heuristics

tanyongkiam:
- faster index unions for Dom
- change strengthen-to-core check using derived set


tanyongkiam:
Tested locally.

PR #1436, opened 2026-07-22
CP encoder

tanyongkiam:
adds a verified CP frontend for the CakePB example.


tanyongkiam:
This was tested locally.

PR #1393, opened 2026-05-19
Pancake Named Shapes

talsewell:
Add the "named shapes" feature to pancake, as implemented by @halogentlepersuasion and subsequently verified by @talsewell .

This permits structs (values of compound shape) to have fields named "x", "y", "head", "tail" etc as opposed to just 0, 1, 2, 3.

The verification and translation is believed to be complete.

PR #1438, opened 2026-07-27
pancake: fix incorrect precedence of not and field accesses

Liamolucko:
Previously, expressions like `!x.0` would be parsed as `(!x).0` instead of `!(x.0)`, leading to shape checking errors because you can't negate anything other than a single word. This PR fixes that by correcting their precedence.

This seems to have been a regression in #1393.


tanyongkiam:
I'll run this locally

PR #1448, opened 2026-08-09
Add more details to Candle soundness

tanyongkiam:
This extends the Candle machine code soundness result.

Namely:

1) it is a bit more precise about what `ctxt` is (an extension of `init_ctxt`) and

2) for the specific HOL light shape of the `ctxt` as in `measured_prefix`, it also gives the consistency result.

### DIFF
--- a/candle/prover/candle_kernel_funsScript.sml
+++ b/candle/prover/candle_kernel_funsScript.sml
@@ -1677,6 +1677,8 @@ Proof
     \\ conj_tac
     >- (
       first_assum $ irule_at Any
+      \\ reverse conj_tac
+      >- fs[STATE_def, CONTEXT_def]
       \\ simp[thm2bytes_def, MAP_MAP_o, o_DEF]
       \\ AP_TERM_TAC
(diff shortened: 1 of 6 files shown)


xrchz (approved):
seems reasonable to me. I didn't look that closely at the new theory... the definitions that appear in the public theorems used outside seem reasonable though

tanyongkiam:
> seems reasonable to me. I didn't look that closely at the new theory... the definitions that appear in the public theorems used outside seem reasonable though

The new theory supplies the new `hol_light_ctxt` relation which is used in the public theorems.

In addition, it also proves this theorem:

```
Theorem measured_prefix_hol_light_ctxt:
  ∀l. axiom_free l ⇒ hol_light_ctxt (l ++ measured_prefix)
```

where `measured_prefix` is everything in context if you run the standard HOL light / Candle path up to the axiom of infinity.

myreen (approved):

tanyongkiam:
Tested locally.

PR #1458, opened 2026-08-16
Fix interference oracles in FFI and ccache calls

tanyongkiam:
This changes `installed` to underspecify FP regs and non-callee-saved registers (they were prescribed too strictly before).


myreen (approved):

tanyongkiam:
Will run locally

PR #1465, opened 2026-08-20
Mcandidate fix

tanyongkiam:


examples/deflate/translation/Holmakefile
mn200, line 1: Feels like this (and other similar inclusions) could move up to top-level holproject.toml file.

examples/sat_encodings/translation/Holmakefile
mn200, line 1: Second occurrence of bootstrap; definitely deserves promotion :)

tutorial/problems/Holmakefile
mn200, line 1: Seems like this could also be a holproject.toml governed sub-tree, even if the top-level excludes it from normal "all-of-cakeml" builds.

examples/deflate/translation/Holmakefile
tanyongkiam: Done — promoted `$(HOLDIR)/examples/bootstrap` to `external_includes` in the top-level `holproject.toml`, and applied the same treatment to every other `$(HOLDIR)` include used by more than one directory:

- `examples/bootstrap` — was duplicated across 4 dirs (this one, `sat_encodings/translation`, `cv_translator`, `examples`)
- `l3-machine-code/{arm,arm8,mips,riscv,x64}/model` — 3 dirs each
- `l3-machine-code/common` — 8 dirs

That removed the redundant lines from 13 Holmakefiles. I left the single-use includes local (`Crypto*`, `miller/miller`, `machine-code/decompiler`, the per-arch `step`/`prog` dirs, `A64_ISA_v86A`, `pl-semantics/lprefix_lub`), on the grounds that promoting an include only one directory uses just moves the coupling rather than removing it — happy to promote those too if you'd rather have all `$(HOLDIR)` references in one place.

examples/sat_encodings/translation/Holmakefile
tanyongkiam: Promoted — and it was worse than two occurrences: `bootstrap` appeared in 4 Holmakefiles. All four now inherit it from `external_includes`. Same for the per-arch `model` dirs (3 each) and `l3-machine-code/common` (8).

tutorial/problems/Holmakefile
tanyongkiam: I tried this and it breaks the root build, so I've left the directory as a plain exclusion for now.

Adding `tutorial/problems/holproject.toml` (with a `[projects.cakeml]` back-link) causes `Holmake -r` at the root to plan 51 targets in there — `arith_exp_demoTheory`, `simple_bstTheory`, … — even though `tutorial/problems` is listed in the root project's `exclude`. Those are the generated exercise files with the solutions stripped out, so they can't compile.

The behaviour looks asymmetric: `exclude` hides project roots *strictly beneath* the excluded directory, but a `holproject.toml` **at** the excluded path is still discovered and joins `-r` unconditionally. Concretely, before this PR `tutorial/solutions` was invisible to root `-r` precisely because it sat under an excluded `tutorial` — that's what motivated splitting the tree into `problems/` and `solutions/` in the first place. I also tried making `tutorial` itself a project with `exclude = ["problems"]`, and `problems` was still built (as `$(tutorial)/problems`).

`holmake = false` isn't a way out either, since it makes the project-mode keys inert and so removes the dependency resolution that would be the point of having the file.

Is that discovery-vs-exclude interaction intended? I've only observed it rather than read the discovery walk, so I may have mis-set something up. If a project root at an excluded path should be skipped by `-r`, then this comment becomes straightforward to implement and I'm happy to revisit. Otherwise `tutorial/problems` has to keep its own `INCLUDES` (which is also why its `balanced_bst` entry stays local — nothing outside the project inherits `external_includes`).

myreen (approved):

Issue #1469, opened 2026-08-24
Link binary with (recently added) features - Pancake


mktnk3:
based on some feedback from the local Pancake users:

As more features are added, we often find ourselves in a situation where we have to figure out what features are supported by a particular binary we have at hand. This is certainly possible via some chasing, but it would be nice to have more direct ways to obtain such information, e.g.:

- add links to the corresponding PRs/commits/binary for each entry in `pancake/NEWS.md` (which connects human readable form to binary)
- add an option to binary that prints out the most recently added features (or even just the content of `NEWS.md`, up to the newly added features)

IlmariReissumies:
The second option sounds easy to do, I'll take a stab at it :)

KurtWu10:
The motivation for the issue is to improve the compatibility of Pancake programs, i.e. to enable programmers to write a single Pancake program that selects language features based on availability, similar to feature test macros in C.

For example, future versions of the compiler may support the bitwise negation operator. However, there is no easy way to write a single Pancake program text (with C macros) that uses the operator on newer compilers while using alternative implementations on older compilers.

KurtWu10:
After discussing with the systems people (Julia @midnightveil), we decided that we will use the latest Pancake compiler for Microkit components for simplicity, and there may not be a need for such a feature.

midnightveil:
Having such a feature where we can print a version is likely a good idea - we can at least provide a helpful error in our build system.

In theory also supporting a way to test for new features seems useful, but it's not critical at this stage (& for the amount of time we have to care about backcompat).

KurtWu10:
currently we can print the version of the compiler

```
$ cake --version
The CakeML compiler

Version details:
Mon Aug 10 15:23:36 2026 UTC

CakeML: 2a49ebcb2da1d79e8d54bf130c203425d93e03e6
HOL4:   fac52534ceb43806d35b11f91dafd558266bcbb6
PolyML: Poly/ML 5.9 Release    RTS version: X86_64-5.9
```

IlmariReissumies:
> In theory also supporting a way to test for new features seems useful, but it's not critical at this stage (& for the amount of time we have to care about backcompat).

Okay good to know @midnightveil ! Even if it's not urgent, at this point I'll finish the PR anyway since I'm almost done with it :)

As a side note, it's not entirely obvious to me that you'd always be able to use the latest version. For example, if some other tooling you depend on (such as the Pancake to Viper pipeline) falls behind mainline Pancake a bit.

[...]

PR #1470, opened 2026-08-24
Add a command line option to print the Pancake news file

IlmariReissumies:
The flag is `--pancake_news`. Addresses #1469


mktnk3:
I suppose this assumes that `NEWS.md` is already updated for the feature being added. I wonder if there is a way to ensure this is done. For example, requiring a non-empty (markdown?) string describing the changes that gets concatenated to the content of `NEWS.md` for both command line printing and updating the file... won't work probably :D Not sure how to go about it, but maybe it's worth stating that "`NEWS.md` should be updated" somewhere...

IlmariReissumies:
Yes that's a good point, this works so long as we're disciplined about updating `NEWS.md` when merging Pancake PRs with user-facing changes. But it doesn't do anything to enforce this discipline.

I think that's difficult to enforce entirely automatically, because there's no good check for figuring out which changes are user-facing. But there are things that would probably help. Maybe a github hook that checks if PRs make any changes in the `pancake/` directory and encourages `NEWS.md` updates or makes you tick a box that it's not needed?

Even so I guess this at least gives people a lower bound on the features they have :)

KurtWu10:
Is there a simple way to make it more machine-readable?

The motivation for the issue is to improve the compatibility of Pancake programs, i.e. to enable programmers to write a single Pancake program that selects language features based on availability, similar to feature test macros in C. In the existing form, I guess I would write another wrapper to retrieve the latest date in the news, and then use the date to query a pre-defined list of dates to know the supported features.

mktnk3:
> Even so I guess this at least gives people a lower bound on the features they have :)

I can't quite decide whether that is a plus or a potential cause for confusion, but either way, that only won't be good enough to close the issue. I guess we can keep looking into more complete solutions.

@KurtWu10 would you like to add something on the motivation to the issue #1469 ?

IlmariReissumies:
> Is there a simple way to make it more machine-readable?

Don't see why not. For example, the `NEWS.md` could assign a tag to each new feature, say `varshifts` for variable-length shifts. You could then ask the binary whether certain tags are present with (say) `cake --pancake_news varshifts`, which would print `true` or `false` accordingly. Would that work for you?

KurtWu10:
Yes, that would work.

IlmariReissumies:
Okay, that's added now! Please check if it looks good to you :)

I added annotations for enabling and disabling features to `NEWS.md`, see here: https://github.com/CakeML/cakeml/blob/pan_news/pancake/NEWS.md

`newsScript` now parses these tags does some basic well-formedness checks on them. The compiler can query them with e.g. `cake --pancake_feature=varshifts`, which will print `true` if the most recent `varshift` entry in `NEWS.md` says "Feature enabled".

KurtWu10:
Thanks for the work, it looks helpful. I noticed that some important features are not in the `NEWS.md` file, e.g. function inlining of PR 1323.

IlmariReissumies:
Good catch! I added one retroactively.

IlmariReissumies:
In the most recent commit I dropped the `--pancake_news` option, since @myreen pointed out that it's inadvisable to have huge one-time-use strings in memory. Now the PR only includes the option to query feature tags with `--pancake_feature=<feature tag>`. This seems closer to the spirit of #1469 and @KurtWu10's ideas anyway :)

PR #1478, opened 2026-09-01
Remove duplicate mllist mapPartial definition

### LINKED ISSUE #1449: Remove mapPartial from mllist
It seems that mllist redefines mapPartial which already exists in list:
```
listTheory.mapPartial_def;
val it =
   ⊢ (∀f. mapPartial f [] = []) ∧
     ∀f x xs.
       mapPartial f (x::xs) =
       case f x of NONE => mapPartial f xs | SOME y => y::mapPartial f xs:
   thm
> mllistTheory.mapPartial_def;
val it =
   ⊢ (∀f. mapPartial f [] = []) ∧
     ∀f h t.
       mapPartial f (h::t) =
       case f h of NONE => mapPartial f t | SOME x => x::mapPartial f t: thm
```
I ran into this issue after spending some time not understanding why `mapPartial_def` is not doing anything as I had an `mllist` as my ancestors.

I suspect the fix is basically as simple as just deleting the definition, with perhaps minor changes required here and there.

mah1104ahm:
## Summary

- remove the duplicate `mapPartial` definition from `mllist`
- prove the existing `mapPartial_thm` against `listTheory.mapPartial_def`
- leave the public theorem available to downstream scripts

## Verification

- confirmed `basis/pure/mllistScript.sml` no longer defines `mapPartial`
- `git diff --check`
- full HOL build not run locally because `Holmake` is unavailable

Closes #1449

### DIFF
--- a/basis/pure/mllistScript.sml
+++ b/basis/pure/mllistScript.sml
@@ -384,17 +384,10 @@ Proof
   |> SIMP_RULE (srw_ss()++ETA_ss) [])]
 QED
 
-Definition mapPartial_def:
-  (mapPartial f [] = []) /\
-  (mapPartial f (h::t) = case (f h) of
-    NONE => mapPartial f t
-    |(SOME x) => x::mapPartial f t)
-End
-
 Theorem mapPartial_thm:
    !f l. mapPartial f l = MAP THE (FILTER IS_SOME (MAP f l))
 Proof
-  Induct_on `l` \\ rw [mapPartial_def] \\ Cases_on `f h` \\ rw [THE_DEF] \\ fs [IS_SOME_DEF]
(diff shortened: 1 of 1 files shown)


tanyongkiam:
Hi, thanks for the contribution.

I think the PR should be done more carefully and with HOL used (at least checking the files that were touched).

For example, there still remains references to mllistTheory.mapPartial_def.

mah1104ahm:
Thank you for the review. You are right: the original change was not
sufficiently verified with HOL.

I found and corrected the three remaining mllistTheory.mapPartial_def
references in holModelConservativityScript.sml locally. I also
inspected regression job 3495: it fails earlier in holExtensionTheory
at abs_or_rep_matches_type_matches, so fixing those references alone
is not enough to claim a working patch.

I do not currently have a working HOL/PolyML environment available
locally. I am holding the follow-up push until the affected theories
can be checked, rather than presenting the textual corrections as
proof validation. Thank you for identifying the gap.

M Ahmed

PR #1482, opened 2026-09-09
Add support for open and let open (CakeML only)

tanyongkiam:
OCaml/Candle syntax support TBD.

Done by Codex (extracted from https://github.com/JUrban ongoing work, which is also Codex)


myreen (approved):
A welcome addition!

semantics/gramScript.sml
```diff
@@ -83,6 +84,22 @@ Definition validPrefixSym_def:
   validPrefixSym s ⇔ s = "~" ∨ "!" ≼ s ∨ "?" ≼ s
 End
 
+(* Qualified identifiers bypass keyword classification in the lexer. Each
+   component of an opened module path must classify as a structure name. *)
+Definition validModName_def:
+  validModName s ⇔ lexer_fun$get_token (explode s) = AlphaT s
+End
+
```
mn200, line 92: Looks to me as if this allows the empty string as a valid module name.  Lower down it looks as if the LLM has re-invented string lists as a type with constructors `End` and `Mod`, and then reinvented `EVERY` over that type (`validModPath`).

mn200 (commented):
Only a minor criticism of the grammar-handling side of things.

semantics/gramScript.sml
myreen: Ah, the `End` and `Mod` list is my manual slop from before LLMs were involved, I think.

tanyongkiam: For the first point, the lexer itself (`get_token`) will fail on empty strings (it's not possible to return AlphaT unless it is a non-empty identifier)

PR #1486, opened 2026-09-12
Change the behavior of Install in preparation for ARM8 support

tanyongkiam:
As the title says, this removes the old behavior of "clear icache" to one that additionally memcopies the installed code directly.

The changeset here gets x64 working again.

A subsequent change will get arm8 + W^X working.


myreen (approved):
Very nice!

PR #1492, opened 2026-09-18
The Pancake decompiler: DecompIt

daole2904:
The initial version of Pancake decompiler into interaction trees, namely DecompIt. The tool comes with a set of examples. There are several issues that I would like to create and address in the future after this initial version:
- Replace usage of `srw_ss` (medium effort, somewhat low reward).
- Rewrite the theorems in `panItreeAbsTheory` to a higher level that does not expose the sum type of the return node, then update the decompiler to use them (medium effort, medium to high reward, currently in progress).


daole2904:
@mn200 @IlmariReissumies I think the decompiler is good to go for the initial version as we discussed!

myreen:
@daole2904 Nice work! This deserves a mention in [changes-since-release.md](https://github.com/CakeML/cakeml/blob/master/developers/changes-since-release.md) and [NEWS.md](https://github.com/CakeML/cakeml/blob/master/pancake/NEWS.md)

